PR #9, opened 2023-08-23
refactor flow for test

jobez:
Time spent on this PR: 1 beautiful day

Resolves: #<Issue number>

# Pull Request type

-   [ ] Bugfix
-   [ ] Feature
-   [X] Refactoring (no functional changes, no API changes)
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?

- Differentiate the `handle` phase of a step from its inner logic, to enable inner logic reuse for different kinds of tests
- Refactor general flow into handle phases

-   [ ] Yes
-   [X] No


crates/ef-testing/src/models/case.rs
```diff
@@ -100,35 +207,11 @@ impl Case for BlockchainTestCase {
...
-                kakarot.proxy_class_hash,
-                eoa_class_hash,
-                kakarot.contract_account_class_hash,
-            );
-            write_test_state(&test, kakarot_address, class_hashes, &mut starknet)
-        })
-        .await
-        .map_err(|err| ef_tests::Error::Assertion(err.to_string()))?;
+        // handle pretest
+        let _handled_pre_state = self.handle_pre_state(&env).await?;
```
jobez, line 211: some notes: 

using `let _` bindings, even when you end a call with the `?`, will result in the error messages not getting conveyed up

jobez (commented):
some justifications/motivations for changes

crates/ef-testing/src/lib.rs
```diff
@@ -40,196 +40,3 @@ pub fn get_signed_rlp_encoded_transaction(
 
     Ok(out.to_vec().into())
 }
-
-#[cfg(test)]
-mod tests {
```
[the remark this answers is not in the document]
greged93: that works for me!

crates/ef-testing/src/models/case.rs
greged93: I tried running this and it does panick in the main so this looks ok
```
fn foo() -> Result<(), std::io::Error> {
    Err(std::io::Error::new(std::io::ErrorKind::Other, "oh no!"))
}

fn bar() -> Result<(), std::io::Error> {
    let _ = foo()?;
    Ok(())
}

fn main() {
    bar().unwrap()
}
```

crates/ef-testing/src/utils/assert.rs
```diff
@@ -0,0 +1,80 @@
...
+) -> Result<(), ef_tests::Error> {
+    let Nonce(actual_nonce) = actual_state.nonce;
+    let account_nonce: FieldElement = Felt252Wrapper::try_from(expected_state.nonce.0)
+        .unwrap()
+        .into();
+
+    let _expected_account_balance: FieldElement =
+        Felt252Wrapper::try_from(expected_state.balance.0)
+            .unwrap()
+            .into();
```
greged93: Propose to remove this for now and add a comment to add it

greged93 (approved):
Made some smol changes to be able to branch out of main, if you don't agree with them we can update later!

PR #18, opened 2023-08-24
feat: improve test running

greged93:
- Allows to run a specific test file or set of tests with regexp.
- Updates the tracing to avoid logging `jsonrpsee`.
- Corrects the assertion display to use the same type, hence displaying both values in hex.

Time spent on this PR: 0.3 days

Resolves: #NA

# Pull Request type

-   [ ] Bugfix
-   [ ] Feature
-   [x] Refactoring (no functional changes, no API changes)
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
Cleans the current use of the tests

-   [ ] Yes
-   [x] No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,17 +7,22 @@ EF_TESTS_DIR := ./crates/ef-testing/ethereum-tests
 
 # Downloads and unpacks Ethereum Foundation tests in the `$(EF_TESTS_DIR)` directory.
 # Requires `wget` and `tar`
+.PHONY:$(EF_TESTS_DIR)
 $(EF_TESTS_DIR):
-	mkdir $(EF_TESTS_DIR)
+	mkdir -p $(EF_TESTS_DIR)
 	wget $(EF_TESTS_URL) -O ethereum-tests.tar.gz
 	tar -xzf ethereum-tests.tar.gz --strip-components=1 -C $(EF_TESTS_DIR)
 	rm ethereum-tests.tar.gz
 
-ef-tests: ef-tests-run ef-tests-clean
+setup: $(EF_TESTS_DIR)
 
-ef-tests-run: $(EF_TESTS_DIR)
+# Runs the Ethereum Foundation tests
+ef-tests:
 	cargo nextest run -p ef-testing --features ef-tests 
 
-ef-tests-clean: 
-	rm -fr $(EF_TESTS_DIR)
+# Runs specific test
+ef-test-single:
+	RUN_SPECIFIC_TESTS=$(test) cargo test -p ef-testing --features ef-tests -- --nocapture
 
+ef-test-file:
+	RUN_SPECIFIC_FILES=$(file) cargo test -p ef-testing --features ef-tests -- --nocapture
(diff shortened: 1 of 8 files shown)


crates/ef-testing/src/models/case.rs
```diff
@@ -216,6 +216,9 @@ impl Case for BlockchainTestCase {
         let specific_tests_to_run: Option<HashSet<String>> = std::env::var("RUN_SPECIFIC_TESTS")
             .ok()
             .map(|test_names| test_names.split(',').map(String::from).collect());
+        let specific_file_to_run: Option<HashSet<String>> = std::env::var("RUN_SPECIFIC_FILES")
```
jobez, line 219: an idea of how we can achieve the same thing with a single environment variable, i.e. having the ability to do

```
RUN_SPECIFIC_TESTS=jumpi_d14g0v0_Shanghai cargo test -p ef-testing --features ef-tests -- --nocapture
```

or 

```
RUN_SPECIFIC_TESTS=jumpi* cargo test -p ef-testing --features ef-tests -- --nocapture

```

- instead of parsing one or many strings, we treat input as regex
- instead of checking if a string is in a set, we check if the test_name matches the regex

wdyt?

jobez (changes requested):
an idea to consolidate functionality behind one environment variable. can be done in another PR if you feel this one is pressing, np.

crates/ef-testing/src/models/case.rs
greged93: Let me try this!

Makefile
```diff
@@ -7,17 +7,22 @@ EF_TESTS_DIR := ./crates/ef-testing/ethereum-tests
 
 # Downloads and unpacks Ethereum Foundation tests in the `$(EF_TESTS_DIR)` directory.
 # Requires `wget` and `tar`
+.PHONY:$(EF_TESTS_DIR)
```
ClementWalter, line 10: do we really want the `.PHONY` ? see https://hackernoon.com/what-does-phony-do-in-a-makefile

crates/ef-testing/src/models/case.rs
ClementWalter: and maybe then only `TEST=jumpi ...` or `TARGET=jumpi ...` instead of the long `RUN_SPECIFIC_TEST`

crates/ef-testing/src/utils/assert.rs
```diff
@@ -25,7 +25,8 @@ pub fn assert_contract_post_state(
     // TODO: find way to assert on balance
     // assert_eq!(actual_account_balance, StarkFelt::from(expected_account_balance));
 
-    if actual_nonce != StarkFelt::from(account_nonce) {
+    let account_nonce = StarkFelt::from(account_nonce);
```
ClementWalter, line 28: does this change something?

crates/ef-testing/src/utils/assert.rs
```diff
@@ -56,7 +57,8 @@ pub fn assert_contract_post_storage(
                 .copied()
                 .unwrap_or_default();
 
-            if actual_state_value != StarkFelt::from(value) {
+            let value = StarkFelt::from(value);
```
ClementWalter, line 60: does this change something?

greged93: Just allows to have the same type and output both values as hex when the comparison fails

Makefile
greged93: Indeed will remove

jobez (approved):
Lgtm

PR #21, opened 2023-08-25
dev: add `EF_TESTS_DIR` as target

ftupas:
Time spent on this PR: 0

Resolves:

# Pull Request type

-   [X] Bugfix
-   [ ] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?

`make setup` now works for fresh install

-   [ ] Yes
-   [X] No


jobez:
@greged93 you had this originally and it was removed because of a comment right?

greged93:
Indeed but I think we can add it back as now it only runs in setup

greged93 (approved):

Issue #23, opened 2023-08-28
dev: add `should_skip` function


greged93:
Add a function which allows to set the tests we want to skip. See the [implementation](https://github.com/paradigmxyz/reth/blob/main/testing/ef-tests/src/cases/blockchain_test.rs#L137) from Reth.

PR #29, opened 2023-08-28
dev: add function to allow skipping tests

bajpai244:
Time spent on this PR:

Resolves: #23 

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?

There is a `should_skip()` function which can be called with a path, and it can return whether a test should be ignored or not.

-   [ ] Yes
-   [x] No

### DIFF
--- a/crates/ef-testing/src/lib.rs
+++ b/crates/ef-testing/src/lib.rs
@@ -3,6 +3,8 @@ pub mod storage;
 pub mod traits;
 pub mod utils;
 
+use std::path::Path;
+
 use bytes::BytesMut;
 use kakarot_rpc_core::client::constants::CHAIN_ID;
 use reth_primitives::{sign_message, Bytes, SealedBlock, Signature, Transaction};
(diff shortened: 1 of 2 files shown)


crates/ef-testing/src/lib.rs
```diff
@@ -39,3 +41,83 @@ pub fn get_signed_rlp_encoded_transaction(
...
+    }
+
+    matches!(
+        name,
+        // A case can be added like this
+        // | "testFileName.json"
+        // using "" as a placeholder till we get our first filename to ignore
+        | ""
+    ) || path_contains(
+        path_str,
```
bajpai244, line 68: List of directories to ignore.

crates/ef-testing/src/lib.rs
```diff
@@ -39,3 +41,83 @@ pub fn get_signed_rlp_encoded_transaction(
...
     Ok(out.to_vec().into())
 }
+
+/// returns whether a given test at a given path should be ignored or not
+pub fn should_skip(path: &Path) -> bool {
+    let path_str = path.to_str().expect("Path is not valid UTF-8");
+    let name = path.file_name().unwrap().to_str().unwrap();
+
+    // list of files that we do want to include even though their parent directories might be marked for skip
+    let include_files = [
```
bajpai244, line 51: Since `BlockchainTests`s are skipped, a special array of files can be considered as an exception.

When a filepath matches them, we ignore the skip test as we know they should not be skipped.

Issue #41, opened 2023-08-30
Fix handling pre-state to create contract accounts when account has storage


ftupas:
In creating the eoa accounts in pre-state, we only check if it has no code. We need to update this so that we also check if it has no storage as well

https://github.com/kkrt-labs/ef-tests/blob/8d64b0a14b68b5dd5e14ada8e8700b692967e132/crates/ef-testing/src/storage/mod.rs#L84-L89

PR #42, opened 2023-08-30
dev: fix handling pre-state to use correct implementation class hash

ftupas:
Time spent on this PR: 0.1

Resolves: #41 

# Pull Request type

-   [X] Bugfix
-   [ ] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?

For EOA accounts, we both check that it has no code nor storage

-   [ ] Yes
-   [X] No

### DIFF
--- a/crates/ef-testing/src/storage/mod.rs
+++ b/crates/ef-testing/src/storage/mod.rs
@@ -1,6 +1,7 @@
 pub mod contract;
 pub mod eoa;
 
+use ef_tests::models::Account;
 use ef_tests::models::State;
 use hive_utils::{
     kakarot::compute_starknet_address,
@@ -84,9 +85,9 @@ pub fn write_test_state(
             })
             .collect::<Result<Vec<()>, RunnerError>>()?;
 
-        let proxy_implementation_class_hash = if account_info.code.is_empty() {
+        // Initialize the EOA or contract account
+        if account_info.code.is_empty() && account_info.storage.is_empty() {
             initialize_eoa(kakarot_address, address, &mut storage)?;
-            class_hashes.eoa_class_hash
         } else {
             initialize_contract_account(
                 kakarot_address,
@@ -95,8 +96,11 @@ pub fn write_test_state(
                 &account_info.code,
                 &mut storage,
             )?;
-            class_hashes.contract_account_class_hash
         };
+
+        // get the implementation class hash
+        let proxy_implementation_class_hash =
+            implementation_class_hash(account_info, &class_hashes);
 
         // write implementation state of proxy
         let proxy_implementation_storage_tuples = genesis_set_storage_starknet_contract(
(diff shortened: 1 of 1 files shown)


crates/ef-testing/src/storage/mod.rs
```diff
@@ -228,3 +231,67 @@ pub fn read_balance(
         .get(&storage_key)
         .unwrap())
 }
+
+/// Check what the implementation class hash should be.
+pub fn implementation_class_hash(
+    account_info: &Account,
+    class_hashes: &ClassHashes,
+) -> FieldElement {
+    if account_info.code.is_empty() && account_info.storage.is_empty() {
```
ftupas, line 240: here now we check that account has no storage as well for it to be an EOA

crates/ef-testing/src/storage/mod.rs
```diff
@@ -228,3 +231,67 @@ pub fn read_balance(
...
+    }
+
+    #[rstest]
+    #[case(r#"{
+        "balance" : "0x00",
+        "code" : "0x",
+        "nonce" : "0x00",
+        "storage" : {
+            "0x01" : "0x01"
+        }
```
ftupas, line 268: CA no code but with storage, example taken from [here](https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/stSStoreTest/InitCollision.json#L684-L691)

crates/ef-testing/src/storage/mod.rs
ClementWalter: so according to my testing on goerli, it's possible to have an CA with no bytecode.

So a CA can have no bytecode and no storage, but it can also have no bytecode and a storage (the initi bytecode returns an empty bytes but set some storage slots).

But an empty CA cannot make any update to its storage nor its nonce, so we probably need to check the post state as well to make sure it's CA or EOA

ClementWalter: tl;dr is: if it has no code and no storage and nonce = 0, in pre and post than we cannot know if it's an EOA or CA, but it probably doesn't matter

ClementWalter (changes requested):
can you just add a comment to the function (this https://github.com/kkrt-labs/ef-tests/pull/42/files#r1310109420) so to make sure that we understand later one the condition?

crates/ef-testing/src/storage/mod.rs
greged93: I am not sure I get the point of this function, can't you just do just as before? Return `class_hashes.eoa_class_hash` or `class_hashes.contract_account_class_hash` in the `if` on line 85?

greged93 (commented):
just one comment

crates/ef-testing/src/storage/mod.rs
ftupas: I wanted to do this at first but decided to make it into a function so I can do a unit test, I do agree it's not idiomatic so open to updating this, what do you think? @ClementWalter @greged93

greged93: I just find it a little weird because it means you branch twice in your code on `if account_info.code.is_empty() && account_info.storage.is_empty() `: once in the function and once before. This just add boiler plate. You could maybe update the function to return `true` or `false` (so like `is_eoa_contract`) and then branch on this?

ftupas: Gotcha yeah I think that's the way to do it, will update accordingly, thanks @greged93 !

ClementWalter: I don't really like the function as it is now, the `class_hashes` arguments are transparent to the logic.

According to its logic, its core goal is to determine wether or not an account is a eoa. For this, it needs `account_info` only.

I think that you could just avec written a `fn is_account_eoa` and replace 

```
let proxy_implementation_class_hash = if account_info.code.is_empty() {
```

by 

```
let proxy_implementation_class_hash = if is_account_eoa(account_info) {
```

ClementWalter (changes requested):
plz update the util to return only true/false

crates/ef-testing/src/storage/mod.rs
ftupas: Gotcha that's much cleaner, I've simplified it 🙏 sorry for the roundabout

ClementWalter (approved):

Issue #48, opened 2023-08-31
bug: init pre state misses values in runner


greged93:
# Bug Report

**Current behavior:**
The runner does not initialize in the pre state the `evm_address` starknet storage of a contract nor the `is_initialized_` storage value for both an eoa or a contract.

**Expected behavior:**
The above mentionned starknet storage values should be initialized in the pre state.

PR #55, opened 2023-09-01
fix: pre state

greged93:
Fixes the pre state of the eoa and contract accounts by adding `evm_address` and `is_initialized`.

Time spent on this PR: 0.1 day

Resolves: #48 

# Pull Request type

-   [x] Bugfix
-   [ ] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
The pre state of the runner is correctly set.

-   [ ] Yes
-   [x] No


ClementWalter (approved):
I'd like us to remove all `mut` parameters I'm discovering in here. We should avoid to mute function parameters as much as possible (ie, never do it plz, unless very strong reason we could discuss)

greged93:
I'm not sure how to mutate the storage, what is the issue with having mut input to a function?

Issue #64, opened 2023-09-04
dev: refactor tests utils to remove all mutable parameters


ClementWalter:
## Context

In few places (I didn't do an exhaustive search, but at least in [crates/ef-testing/src/storage/mod.rs](../blob/crates/ef-testing/src/storage/mod.rs)) we `mut` an input parameters.

We should not do this. Google search "why mutating input arg is wrong" will give some clues.

## Task

Go though the whole codebase and rewrite everything as required to remove all these `mut`.

PR #69, opened 2023-09-05
dev: remove mut

greged93:
Remove mutables as inputs of functions

Time spent on this PR: 0.4 day

Resolves: #64 

# Pull Request type

-   [ ] Bugfix
-   [ ] Feature
-   [x] Refactoring (no functional changes, no API changes)
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
Mutable inputs have been removed from the codebase (except for `write_test_state` where the `RwLock` needs to be mutable).

-   [ ] Yes
-   [x] No

### DIFF
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -3,7 +3,9 @@
 use super::{error::RunnerError, result::CaseResult, BlockchainTestTransaction};
 use crate::{
     get_signed_rlp_encoded_transaction,
-    storage::{eoa::get_eoa_class_hash, read_balance, write_test_state, ClassHashes},
+    storage::{
+        eoa::get_eoa_class_hash, fee_token::read_balance, models::ClassHashes, write_test_state,
+    },
     traits::Case,
     utils::{
         assert::{assert_contract_post_state, assert_empty_post_state},
@@ -130,7 +132,9 @@ impl BlockchainTestCase {
         let post_state = match test.post_state.as_ref().ok_or_else(|| {
             RunnerError::Other(format!("missing post state for {}", test_case_name))
         })? {
-            RootOrState::Root(_) => panic!("RootOrState::Root(_) not supported, for {}", test_case_name),
+            RootOrState::Root(_) => {
+                panic!("RootOrState::Root(_) not supported, for {}", test_case_name)
+            }
             RootOrState::State(state) => state,
         };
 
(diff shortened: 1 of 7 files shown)


crates/ef-testing/src/models/case.rs
ClementWalter, line 1: looks like there is a new for a CI in this repo

crates/ef-testing/src/storage/contract.rs
```diff
@@ -1,61 +1,47 @@
...
-    destination: &mut HashMap<StarknetStorageKey, StarkFelt>,
-) -> Result<(), RunnerError> {
-    let evm_address_key = get_starknet_storage_key("evm_address", &[], 0)?;
-    destination.insert(evm_address_key, Into::<StarkFelt>::into(evm_address));
-    Ok(())
+) -> Result<Vec<(StarknetStorageKey, StarkFelt)>, RunnerError> {
+    let mut contract_storage = vec![
+        get_evm_address(evm_address)?,
+        get_is_initialized()?,
+        get_owner(kakarot_address)?,
```
ClementWalter, line 20: could we rename to

```
evm_address_storage(...
is_initialiazed_storage(...
owner_storage(...
```

(btw I thiink that the `get_` is "banned from rust, cc @Eikix)

crates/ef-testing/src/storage/eoa.rs
```diff
@@ -47,36 +45,22 @@ pub fn get_eoa_class_hash(
...
+        get_kakarot_address(kakarot_address)?,
+        get_evm_address(evm_address)?,
+    ];
+    Ok(eoa_storage)
 }
 
-/// Writes the kakarot address to a hashmap.
-fn write_kakarot_address(
+/// Returns the kakarot address storage tuple.
+fn get_kakarot_address(
```
ClementWalter, line 62: same for the `get_` i'd prefer `_storage`

ClementWalter (commented):
just this comment regarding the `get_` prefix, and proposed `_storage` suffix instead

crates/ef-testing/src/storage/contract.rs
greged93: Yeah indeed decided to add the `get_` since I had the same understanding that `get_` is banned for getters. Can remove but I think `get` adds context in this case.

crates/ef-testing/src/storage/eoa.rs
greged93: got it will update

ClementWalter (approved):

PR #82, opened 2023-09-08
refactor: refactor according to clippy pedantic recommendation and add more documentation

Eikix:
Time spent on this PR: 0.1day

Resolves #81 

# Pull Request type

-   [ ] Bugfix
-   [ ] Feature
-   [ ] Build-related changes
-   [x] Documentation content changes
-   [ ] Testing

# What is the new behavior?

-   [ ] Yes
-   [x] No

### DIFF
--- a/crates/ef-testing/src/bin/fetch_dump_katana.rs
+++ b/crates/ef-testing/src/bin/fetch_dump_katana.rs
@@ -61,7 +61,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         .default_headers(headers)
         .build()?;
     let mut response = client_gzip
-        .get(&latest_artifact.archive_download_url)
+        .get(latest_artifact.archive_download_url)
         .send()?;
 
     let mut out = fs::File::create("temp.zip")?;
--- a/crates/ef-testing/src/bin/fetch_dump_katana.rs
+++ b/crates/ef-testing/src/bin/fetch_dump_katana.rs
@@ -19,39 +19,33 @@ struct WorkflowRun {
     head_branch: String,
 }
 
+/// Fetches the Katana dump from Kakarot-RPC artifacts
+/// On every PR and merge, thanks to a CI job in Kakarot-RPC, the state of a Starknet devnet
+/// with all of Kakarot Cairo smart contracts deployed is dumped in an artifact called 'dump-katana'
+/// Starting a Starknet local chain from checkpoint speeds up all our processes.
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     dotenv().ok();
     let token =
         std::env::var("GITHUB_TOKEN").map_err(|_| eyre::eyre!("Missing GITHUB_TOKEN in .env"))?;
-    let url = "https://api.github.com/repos/sayajin-labs/kakarot-rpc/actions/artifacts";
+    let url = "https://api.github.com/repos/kkrt-labs/kakarot-rpc/actions/artifacts";
 
-    let client = reqwest::blocking::Client::builder();
-    let response: serde_json::Value = client
+    let client = reqwest::blocking::Client::builder()
         .user_agent("reqwest-rust")
-        .build()?
-        .get(url)
-        .send()?
-        .json()?;
+        .build()?;
+    let response: serde_json::Value = client.get(url).send()?.json()?;
 
-    // Filter the artifacts to only include the ones we care about
-    // and sort by updated_at
+    // Filter the artifacts to only include dump-katana artifacts
+    // and find the latest one by key `updated_at`
     let artifacts: Vec<Artifact> = serde_json::from_value(response["artifacts"].clone())?;
-    let mut artifacts = artifacts
+    let latest_artifact = artifacts
         .into_iter()
         .filter(|artifact| {
             artifact.name == "dump-katana" && artifact.workflow_run.head_branch == "main"
         })
-        .collect::<Vec<_>>();
-    artifacts.sort_by_key(|artifact| artifact.updated_at);
-    artifacts.reverse();
-
-    // Find the latest artifact
-    let latest_artifact = artifacts
-        .first()
+        .max_by_key(|artifact| artifact.updated_at)
         .ok_or_else(|| eyre::eyre!("Missing artifact value for Katana dump"))?;
 
     // Download the artifact
-    let client = reqwest::blocking::Client::builder();
     let mut headers = header::HeaderMap::new();
     let mut auth = header::HeaderValue::from_str(&format!("Bearer {}", token))?;
     auth.set_sensitive(true);
@@ -61,11 +55,12 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         header::HeaderValue::from_static("application/vnd.github+json"),
     );
 
-    let mut response = client
-        .user_agent("reqwest-rust")
+    let client_gzip = reqwest::blocking::Client::builder()
+        .user_agent("reqwest-rust-with-gzip")
         .gzip(true)
         .default_headers(headers)
-        .build()?
+        .build()?;
+    let mut response = client_gzip
         .get(&latest_artifact.archive_download_url)
         .send()?;
 
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -53,7 +53,7 @@ async fn handle_pre_state(
         eoa_class_hash,
         kakarot.contract_account_class_hash,
     );
-    write_test_state(pre_state, kakarot_address, &class_hashes, &mut starknet)?;
+    write_test_state(pre_state, kakarot_address, class_hashes, &mut starknet)?;
     Ok(())
 }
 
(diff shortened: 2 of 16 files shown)


crates/ef-testing/src/bin/fetch_dump_katana.rs
```diff
@@ -19,39 +19,33 @@ struct WorkflowRun {
...
+/// Fetches the Katana dump from Kakarot-RPC artifacts
+/// On every PR and merge, thanks to a CI job in Kakarot-RPC, the state of a Starknet devnet
+/// with all of Kakarot Cairo smart contracts deployed is dumped in an artifact called 'dump-katana'
+/// Starting a Starknet local chain from checkpoint speeds up all our processes.
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     dotenv().ok();
     let token =
         std::env::var("GITHUB_TOKEN").map_err(|_| eyre::eyre!("Missing GITHUB_TOKEN in .env"))?;
-    let url = "https://api.github.com/repos/sayajin-labs/kakarot-rpc/actions/artifacts";
+    let url = "https://api.github.com/repos/kkrt-labs/kakarot-rpc/actions/artifacts";
```
greged93, line 30: Are you sure this is correct? [Python script](https://github.com/kkrt-labs/kakarot/blob/main/scripts/artifacts.py#L22C10-L22C77) in `kakarot` repo uses `https://api.github.com/repos/sayajin-labs/kakarot/actions/artifacts`.

crates/ef-testing/src/bin/fetch_dump_katana.rs
```diff
@@ -19,39 +19,33 @@ struct WorkflowRun {
...
             artifact.name == "dump-katana" && artifact.workflow_run.head_branch == "main"
         })
-        .collect::<Vec<_>>();
-    artifacts.sort_by_key(|artifact| artifact.updated_at);
-    artifacts.reverse();
-
-    // Find the latest artifact
-    let latest_artifact = artifacts
-        .first()
+        .max_by_key(|artifact| artifact.updated_at)
```
greged93, line 45: nice

crates/ef-testing/src/bin/fetch_dump_katana.rs
Eikix: ahh, there's a permanent alias between the two, but it's safer to update to kkrt-labs in the python script as well

crates/ef-testing/tests/tests.rs
```diff
@@ -31,12 +31,15 @@ fn setup() {
 }
 
 pub fn verify_kakarot_sha() -> Result<String, eyre::Error> {
+    // This is the SHA hash of the latest Kakarot submodule commit, inside Kakarot-RPC
     let remote_sha = fs::read_to_string("../../.katana/remote_kakarot_sha")?;
+    // This is your local SHA hash of the Kakarot commit you last pulled, using `make fetch-dump`
```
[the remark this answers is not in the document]
Eikix: who would ever want to suffer such pain :(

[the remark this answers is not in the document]
greged93: haha for example (adding this to the CONTRIBUTOR.md) you can build a dump that allows you to display the opcodes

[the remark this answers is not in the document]
Eikix: So I should update this comment?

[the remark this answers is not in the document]
greged93: I would say something like `// This is your local SHA hash of the Kakarot commit currently used in the dump`.

crates/ef-testing/src/models/case.rs
```diff
@@ -184,7 +188,7 @@ impl BlockchainTestCase {
                     continue;
                 }
                 Some(state) => {
-                    assert_contract_post_state(test_case_name, evm_address, expected_state, state)?
+                    assert_contract_post_state(test_case_name, evm_address, expected_state, state)?;
```
jobez, line 191: why are these changes added quietly and in a pr for docs?

jobez (changes requested):
are most of the syntax changes in this PR just running a linter? seems like there are some changes that would be educational to have explained. can you add a comment to the code changes if you think it would be instructive?

by comment, i mean within the PR, not on the code! (would be a good learning opportunity for me, no worries if nothing stands out)

crates/ef-testing/src/bin/fetch_dump_katana.rs
```diff
@@ -19,39 +19,33 @@ struct WorkflowRun {
...
 fn main() -> Result<(), Box<dyn std::error::Error>> {
     dotenv().ok();
     let token =
         std::env::var("GITHUB_TOKEN").map_err(|_| eyre::eyre!("Missing GITHUB_TOKEN in .env"))?;
-    let url = "https://api.github.com/repos/sayajin-labs/kakarot-rpc/actions/artifacts";
+    let url = "https://api.github.com/repos/kkrt-labs/kakarot-rpc/actions/artifacts";
 
-    let client = reqwest::blocking::Client::builder();
-    let response: serde_json::Value = client
+    let client = reqwest::blocking::Client::builder()
```
Eikix, line 32: here refactoring a little because semantically it's only a client once it's built with .build(), otherwise it's still a builder

crates/ef-testing/src/bin/fetch_dump_katana.rs
Eikix, line 45: we can just get the max here

crates/ef-testing/src/models/case.rs
```diff
@@ -52,7 +52,7 @@ async fn handle_pre_state(
         eoa_class_hash,
         kakarot.contract_account_class_hash,
     );
-    write_test_state(pre_state, kakarot_address, class_hashes, &mut starknet)?;
+    write_test_state(pre_state, kakarot_address, &class_hashes, &mut starknet)?;
```
Eikix, line 55: this is clippy saying we're passing by value but not consuming
only by reference is enough,
less is more

crates/ef-testing/src/models/case.rs
```diff
@@ -122,7 +126,7 @@ impl BlockchainTestCase {
         let block = test
             .blocks
             .first()
-            .ok_or(RunnerError::Other("test has no blocks".to_string()))?
+            .ok_or_else(|| RunnerError::Other("test has no blocks".to_string()))?
```
Eikix, line 129: this is clippy saying we should use ok_or_else so that the function within the brackets is lazily evaluated, 
if we use ok_or, it's run no matter what, 

so the takeaway is that you should use ok or else when using functions within the error case

crates/ef-testing/src/models/case.rs
Eikix, line 191: here clippy says we don't return anything in this arm, so it's bad practice to not put a semicolon, since we're returning into the void

crates/ef-testing/src/models/case.rs
```diff
@@ -222,13 +226,13 @@ impl Case for BlockchainTestCase {
         let general_state_tests_path = general_state_tests_path.as_path();
         Ok(Self {
             tests: {
-                let s = load_file(path)?;
-                deserialize_into(s, path)?
+                let file = load_file(path)?;
```
Eikix, line 229: refrain from naming things with one letter it's often confusing

crates/ef-testing/src/models/case.rs
```diff
@@ -257,7 +261,7 @@ impl Case for BlockchainTestCase {
             None => None,
         };
 
-        for (test_name, case) in self.tests.iter() {
+        for (test_name, case) in &self.tests {
```
Eikix, line 264: here clippy wants reference

crates/ef-testing/tests/tests.rs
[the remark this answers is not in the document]
Eikix: done

jobez (approved):

Eikix:
rebased

crates/ef-testing/src/models/suite.rs
```diff
@@ -10,7 +10,8 @@ pub struct BlockchainTestSuite {
 }
 
 impl BlockchainTestSuite {
-    pub fn new(name: String) -> Self {
+    #[must_use]
```
ClementWalter, line 13: why adding this? do you think every `new` should have it?

crates/ef-testing/src/storage/models.rs
```diff
@@ -7,7 +7,8 @@ pub struct ClassHashes {
 }
 
 impl ClassHashes {
-    pub fn new(
+    #[must_use]
+    pub const fn new(
```
ClementWalter, line 11: same regarding `must_use`

ClementWalter (approved):
let's add more and more rules and linters! After trunk is merged on kakarot, we can add it here as well

crates/ef-testing/src/models/suite.rs
Eikix: good question! on this matter:

```
Clippy pedantic will only ask you to add #[must_use] on functions which

- are publicly visible

- have no (internally) mutable arguments

- don't mutate any statics

- return neither ! nor () nor any #[must_use] type
```

Because there is no way to ensure at compile time that this isn't dead code, but with must_use, you gain protection against dead code

PR #100, opened 2023-09-13
feat: bump katana

greged93:
Bump the Katana dep and RPC deps.

Time spent on this PR: 2 days

Resolves: #NA

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
The katana dep and RPC dep is bumped.

-   [ ] Yes
-   [x] No

### DIFF
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -42,23 +42,18 @@ async fn handle_pre_state(
 ) -> Result<(), RunnerError> {
     let kakarot_address = kakarot.kakarot_address;
 
-    let mut starknet_state = env.sequencer().sequencer.backend.state.write().await;
-    let db = starknet_state
+    let mut starknet = env.sequencer().sequencer.backend.state.write().await;
+    let starknet_db = starknet
         .maybe_as_cached_db()
         .ok_or_else(|| RunnerError::Other("failed to get Katana database".to_string()))?;
 
-    let eoa_class_hash = get_eoa_class_hash(env, &db)?;
+    let eoa_class_hash = get_eoa_class_hash(env, &starknet_db)?;
     let class_hashes = ClassHashes::new(
         kakarot.proxy_class_hash,
         eoa_class_hash,
         kakarot.contract_account_class_hash,
     );
-    write_test_state(
-        pre_state,
-        kakarot_address,
-        class_hashes,
-        &mut starknet_state,
-    )?;
+    write_test_state(pre_state, kakarot_address, class_hashes, &mut starknet)?;
     Ok(())
 }
 
@@ -178,10 +173,10 @@ impl BlockchainTestCase {
             let starknet_contract_address =
                 StarknetContractAddress(Into::<StarkFelt>::into(starknet_address).try_into()?);
 
-            let db = starknet
+            let starknet_db = starknet
                 .maybe_as_cached_db()
                 .ok_or_else(|| RunnerError::Other("failed to get Katana database".to_string()))?;
-            let actual_state = db.storage.get(&starknet_contract_address);
+            let actual_state = starknet_db.storage.get(&starknet_contract_address);
             match actual_state {
                 None => {
                     // if no state, check post state is empty
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -15,13 +15,9 @@ use crate::{
 use async_trait::async_trait;
 use ef_tests::models::BlockchainTest;
 use ef_tests::models::{ForkSpec, RootOrState, State};
-use kakarot_rpc_core::{
-    client::api::{KakarotEthApi, KakarotStarknetApi},
-    models::felt::Felt252Wrapper,
-};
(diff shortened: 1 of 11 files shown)


crates/ef-testing/src/models/case.rs
```diff
@@ -168,7 +164,7 @@ impl BlockchainTestCase {
         let kakarot_address = kakarot.kakarot_address;
 
         // Get lock on the Starknet sequencer
-        let starknet = env.sequencer().sequencer.backend.state.read().await;
+        let mut starknet = env.sequencer().sequencer.backend.state.write().await;
```
jobez, line 167: is there a reason why we are calling `write` here?

crates/ef-testing/src/models/case.rs
```diff
@@ -177,11 +173,14 @@ impl BlockchainTestCase {
             let starknet_contract_address =
                 StarknetContractAddress(Into::<StarkFelt>::into(starknet_address).try_into()?);
 
-            let actual_state = starknet.storage.get(&starknet_contract_address);
+            let starknet_db = starknet
```
jobez, line 176: can this be done outside the `for` loop?

crates/ef-testing/src/models/case.rs
```diff
@@ -136,14 +137,9 @@ impl BlockchainTestCase {
...
-
-        // we make sure that the transaction has a receipt and fail fast if it doesn't
-        let starknet_provider = env.client().starknet_provider();
-        let transaction_hash: FieldElement = FieldElement::from_bytes_be(&hash)?;
-        starknet_provider
-            .get_transaction_receipt::<FieldElement>(transaction_hash)
-            .await?;
+        // Send the transaction without checking for errors, accounting
+        // for the fact that some transactions might fail.
+        let _ = client.send_transaction(tx_encoded).await;
```
Eikix, line 142:
```suggestion
        client.send_transaction(tx_encoded).await;
```

Eikix: btw why 
```        // Send the transaction without checking for errors, accounting
```

crates/ef-testing/src/models/case.rs
```diff
@@ -177,11 +173,14 @@ impl BlockchainTestCase {
...
-            let actual_state = starknet.storage.get(&starknet_contract_address);
+            let starknet_db = starknet
+                .maybe_as_cached_db()
+                .ok_or_else(|| RunnerError::Other("failed to get Katana database".to_string()))?;
+            let actual_state = starknet_db.storage.get(&starknet_contract_address);
             match actual_state {
                 None => {
                     // if no state, check post state is empty
-                    let actual_balance = read_balance(evm_address, starknet_address, &starknet)
+                    let actual_balance = read_balance(evm_address, starknet_address, &mut starknet)
```
Eikix, line 183: why does read_balance need a mutable reference to starknet?

jobez (commented):
some smol asks, the finish line approaches 🏃‍♂️

crates/ef-testing/src/storage/mod.rs
```diff
@@ -78,12 +76,15 @@ pub fn write_test_state(
...
-        let storage_record = StorageRecord {
-            nonce: Nonce(StarkFelt::from(account_nonce)),
-            class_hash: ClassHash(class_hashes.proxy_class_hash.into()),
-            storage: starknet_contract_storage.into_iter().collect(),
-        };
-        starknet.storage.insert(address, storage_record);
+        starknet
+            .set_class_hash_at(address, ClassHash(class_hashes.proxy_class_hash.into()))
+            .map_err(|err| {
+                RunnerError::Other(format!("error setting class hash at {address:#?}: {err}"))
```
Eikix, line 82: there are three times RunnerError::Other that are linked to Katana context, should we create a RunnerError::SequencerError or some similar error to express a failure with underlying engine?

Eikix (commented):
great PR! Lfg let's get those ef-tests rekt

crates/ef-testing/src/models/case.rs
greged93: The linter was telling me to add the `let _`, because send_transaction sends back an error which I don't want to handle. This error comes from the fact that the transaction waiter we have in the RPC will send an error if the transaction he is awaiting returns a rejected transaction.

crates/ef-testing/src/models/case.rs
greged93: I was also triggered by this but it comes from the fact that read_balance uses the `Database` from Katana in order to read, and the Database is a super-trait for the `StateReader` from `blockifier`, which takes a mutable reference even for reading.
https://github.com/dojoengine/dojo/blob/main/crates/katana/core/src/fork/db.rs#L100

crates/ef-testing/src/models/case.rs
greged93: See the above comment on the Elias remark.

Eikix (approved):
lgtm

PR #107, opened 2023-09-20
feat: implement blockifier

greged93:
Implement the blockifier State and StateReader for the State structure used in the Kakarot sequencer.

Resolves: #106

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
The state structure that will be used for the test sequencer implements the `State` and `StateReader` traits from blockifier that will be used during the execution of transactions.

-   [ ] Yes
-   [x] No

### DIFF
--- a/crates/sequencer/Cargo.toml
+++ b/crates/sequencer/Cargo.toml
@@ -18,3 +18,6 @@ starknet_api = { workspace = true }
 
 # Other
 rustc-hash = "1.1.0"
+
+[dev-dependencies]
+lazy_static = { workspace = true }
--- a/crates/sequencer/src/state.rs
+++ b/crates/sequencer/src/state.rs
@@ -1,8 +1,14 @@
+use blockifier::state::cached_state::CommitmentStateDiff;
+use blockifier::state::errors::StateError;
+use blockifier::state::state_api::{
+    State as BlockifierState, StateReader as BlockifierStateReader, StateResult,
+};
 use blockifier::{
     execution::contract_class::ContractClass, state::cached_state::ContractStorageKey,
 };
 use rustc_hash::FxHashMap;
 use starknet_api::core::CompiledClassHash;
+use starknet_api::state::StorageKey;
 use starknet_api::{
     core::{ClassHash, ContractAddress, Nonce},
     hash::StarkFelt,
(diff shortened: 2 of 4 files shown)


crates/sequencer/src/state.rs
```diff
@@ -22,3 +28,229 @@ pub struct State {
...
+    }
+
+    fn to_state_diff(&self) -> CommitmentStateDiff {
+        unreachable!("to_state_diff should not be called in the sequencer")
+    }
+}
+
+impl BlockifierStateReader for State {
+    /// Default: 0 for an uninitialized contract address.
+    fn get_storage_at(
```
Eikix, line 93: no need for `get` prefix imo here

crates/sequencer/src/state.rs
```diff
@@ -22,3 +28,229 @@ pub struct State {
...
+    ) -> StateResult<StarkFelt> {
+        Ok(self
+            .storage
+            .get(&(contract_address, key))
+            .cloned()
+            .unwrap_or_default())
+    }
+
+    /// Default: 0 for an uninitialized contract address.
+    fn get_nonce_at(&mut self, contract_address: ContractAddress) -> StateResult<Nonce> {
```
Eikix, line 106: same

crates/sequencer/src/state.rs
```diff
@@ -22,3 +28,229 @@ pub struct State {
...
+    /// Default: 0 (uninitialized class hash) for an uninitialized contract address.
+    fn get_class_hash_at(&mut self, contract_address: ContractAddress) -> StateResult<ClassHash> {
+        Ok(self
+            .contracts
+            .get(&contract_address)
+            .cloned()
+            .unwrap_or_default())
+    }
+
+    /// Errors if the compiled class hash is not declared.
```
Eikix, line 123: you interverted comments i think

Eikix: or at least i'm not sure why the fn is called `get_compiled_contract_class` and we access self.classes

crates/sequencer/src/state.rs
greged93: No choice, this is the blockifier trait

crates/sequencer/src/state.rs
greged93: I think blockifier has this a little reverted, but no choice here again

crates/sequencer/Cargo.toml
```diff
@@ -13,8 +13,12 @@ license.workspace = true
 
 [dependencies]
 # Starknet
-blockifier = { git = "https://github.com/starkware-libs/blockifier.git", tag = "v0.3.0-rc0" }
+blockifier = { package = "blockifier", git = "https://github.com/starkware-libs/blockifier.git", tag = "v0.3.0-rc0" }
```
greged93, line 16: Currently this is being patched by the blockifier patch we have in the workspace manifest. Once we remove the Katana dep, we can remove the patch and this dependency should be applied.

crates/sequencer/src/sequencer.rs
```diff
@@ -1,17 +1,23 @@
...
+where
+    for<'a> &'a mut S: State + StateReader,
+{
+    pub context: BlockContext,
     pub state: S,
 }
 
-impl<S: State + StateReader> Sequencer<S> {
+impl<S> Sequencer<S>
+where
```
Eikix, line 16: We should add more comments on complex rust bounds and generics, i.e.

`For all lifetimes 'a, a mutable reference to S (&'a mut S) must implement both the State and StateReader traits.`

https://doc.rust-lang.org/nomicon/hrtb.html

btw why do we need such a deep magic?

crates/sequencer/src/sequencer.rs
```diff
@@ -1,17 +1,23 @@
...
 use blockifier::state::state_api::{State, StateReader};
 
 /// Sequencer is the main struct of the sequencer crate.
 /// Using a trait bound for the state allows for better
 /// speed, as the type of the state is known at compile time.
-pub struct Sequencer<S: State + StateReader> {
-    pub config: SequencerConfig,
+pub struct Sequencer<S>
+where
+    for<'a> &'a mut S: State + StateReader,
```
Eikix, line 9: same

crates/sequencer/src/sequencer.rs
```diff
@@ -1,17 +1,23 @@
-use crate::config::SequencerConfig;
+use blockifier::block_context::BlockContext;
```
Eikix, line 1: is there a clippy fix to group all imports at once?

crates/sequencer/src/state.rs
```diff
@@ -16,9 +22,235 @@ use starknet_api::{
...
+    classes: FxHashMap<ClassHash, ContractClass>,
+    compiled_class_hashes: FxHashMap<ClassHash, CompiledClassHash>,
+    contracts: FxHashMap<ContractAddress, ClassHash>,
+    storage: FxHashMap<ContractStorageKey, StarkFelt>,
+    nonces: FxHashMap<ContractAddress, Nonce>,
+}
+
+/// State implementation for the sequencer. We use a mutable reference to the state
+/// because this is what will be available during the implementation of the execution.
+impl<'a> BlockifierState for &'a mut State {
```
Eikix, line 34: why do we need lifetimes in our context?

Eikix: we provide lifetime in the impl definition, but never inside the functions below

crates/sequencer/src/state.rs
```diff
@@ -16,9 +22,235 @@ use starknet_api::{
...
+        Ok(())
+    }
+
+    fn to_state_diff(&self) -> CommitmentStateDiff {
+        unreachable!("to_state_diff should not be called in the sequencer")
+    }
+}
+
+impl<'a> BlockifierStateReader for &'a mut State {
+    /// Default: 0 for an uninitialized contract address.
```
Eikix, line 94: This comment isn't so clear
default for what? 

you're saying: storage.get(&(contract_address, key)) will fail if contract_address isn't initialized and thus it'll trigger default case of unwrap_or_default?

crates/sequencer/src/state.rs
```diff
@@ -16,9 +22,235 @@ use starknet_api::{
...
+        &mut self,
+        class_hash: &ClassHash,
+    ) -> StateResult<ContractClass> {
+        self.classes
+            .get(class_hash)
+            .cloned()
+            .ok_or_else(|| StateError::UndeclaredClassHash(class_hash.to_owned()))
+    }
+
+    /// Errors if the compiled class hash is not declared.
```
Eikix, line 136: if you run `cargo clippy --  -W clippy::pedantic -W clippy::nursery` it'll tell you the preferred way to express when a function can error

crates/sequencer/src/state.rs
```diff
@@ -16,9 +22,235 @@ use starknet_api::{
...
+    #[test]
+    #[should_panic(
+        expected = "UndeclaredClassHash(ClassHash(StarkFelt(\"0x0000000000000000000000000000000000000000000000000000000000000001\"))"
+    )]
+    fn test_uninitialized_compiled_class_hash() {
+        // Given
+        let mut state = &mut State::default();
+
+        // When
+        state.get_compiled_class_hash(*ONE_HASH).unwrap();
```
Eikix, line 254: isn't it better to not unwrap and assert on the error? just an idea of preference

crates/sequencer/src/sequencer.rs
greged93: So basically I have to implement State and StateReader not for S but for &mut S, this means that I need to bring a lifetime in, and apparently I need to use for<'a>, I am not a expert on this but I guess it just means that I don't impose restrictions on the lifetime, just saying that for any lifetime, the reference to State must implement State and StateReader (as you said).

crates/sequencer/src/sequencer.rs
greged93: I don't think these should be grouped right? My formatter usually groups them when needed.

crates/sequencer/src/state.rs
greged93: I guess you are right, I don't need the lifetime here, my bad

crates/sequencer/src/state.rs
greged93: Will add `contract address or key`. It just means that if the storage value in the contract is uninitialzed, we return a default of 0.

crates/sequencer/src/state.rs
greged93: This was something that was discussed with @ClementWalter, if I recall correctly he linked a part of the rust book where it indicated you should use #[should_panic]

greged93: It's here, reading through it I am not sure, we could do `is_err()` and then also assert on the specific error with `matches` or something similiar.
https://doc.rust-lang.org/book/ch11-01-writing-tests.html?highlight=should_panic#checking-for-panics-with-should_panic

crates/sequencer/src/state.rs
```diff
@@ -16,9 +22,239 @@ use starknet_api::{
...
+        key: StorageKey,
+    ) -> StateResult<StarkFelt> {
+        Ok(self
+            .storage
+            .get(&(contract_address, key))
+            .cloned()
+            .unwrap_or_default())
+    }
+
+    /// Default: 0 for an uninitialized contract address.
```
Eikix, line 107:
```suggestion
    /// # Default: returns 0 if `contract_address` is not in `State.nonces.keys`
```

crates/sequencer/Cargo.toml
Eikix: Can we add a todo?

crates/sequencer/src/constants.rs
```diff
@@ -0,0 +1,29 @@
...
+        pub static ref TEST_CONTRACT: StarkFelt = *TWO_FELT;
+        pub static ref SENDER_ADDRESS: FieldElement = FieldElement::from(2u8);
+        pub static ref SEQUENCER_ADDRESS: ContractAddress = ContractAddress(TryInto::<PatriciaKey>::try_into(StarkFelt::from(1234u16)).unwrap());
+        pub static ref FEE_TOKEN_ADDRESS: ContractAddress = ContractAddress(TryInto::<PatriciaKey>::try_into(StarkFelt::from(12345u16)).unwrap());
+
+        pub static ref ONE_FELT: StarkFelt = StarkFelt::from(1u8);
+        pub static ref TWO_FELT: StarkFelt = StarkFelt::from(2u8);
+        pub static ref ONE_PATRICIA: PatriciaKey = TryInto::<PatriciaKey>::try_into(*ONE_FELT).unwrap();
+        pub static ref TWO_PATRICIA: PatriciaKey = TryInto::<PatriciaKey>::try_into(*TWO_FELT).unwrap();
+        pub static ref ONE_HASH: ClassHash = ClassHash(*ONE_FELT);
```
Eikix, line 23:
```suggestion
        pub static ref ONE_CLASS_HASH: ClassHash = ClassHash(*ONE_FELT);
```

crates/sequencer/src/constants.rs
```diff
@@ -0,0 +1,29 @@
...
+        pub static ref TWO_HASH: ClassHash = ClassHash(*TWO_FELT);
```
Eikix, line 24:
```suggestion
        pub static ref TWO_CLASS_HASH: ClassHash = ClassHash(*TWO_FELT);
```

crates/sequencer/src/constants.rs
```diff
@@ -0,0 +1,29 @@
...
+        pub static ref ONE_COMPILED_HASH: CompiledClassHash = CompiledClassHash(*ONE_FELT);
```
Eikix, line 25:
```suggestion
        pub static ref ONE_COMPILED_CLASS_HASH: CompiledClassHash = CompiledClassHash(*ONE_FELT);
```

Eikix (approved):
🔥🔥🔥🔥🔥 deez great

Issue #89, opened 2023-09-11
epic: add `stBadOpcode`


Eikix:
# Feature Request

**Describe the Feature Request**

```[tasklist]
### Tasks
- [x] Add the `stBadOpcode` test suite in: https://github.com/kkrt-labs/ef-tests/blob/6b82c723fb123df8ef3dbc9d5efd21843e005a24/crates/ef-testing/tests/tests.rs#L59
- [x] Run the ef tests `make ef-tests`
- [x] See which tests are failing if any, create issues for each unique error message which their test linked (include screenshots, code snippets). Refrain from debugging at this point. This is an integration task.
- [x] If some tests fail, skip them https://github.com/kkrt-labs/ef-tests/blob/6b82c723fb123df8ef3dbc9d5efd21843e005a24/crates/ef-testing/src/models/case.rs#L67
- [x] you're done, now add these issues down below to the linked debuggor issues tasklist for observability
```

```[tasklist]
### Linked debuggor issues
- [x] Add here the 'debuggor' issues resulting from adding the new test suite
- [ ] https://github.com/kkrt-labs/ef-tests/issues/121
- [ ] https://github.com/kkrt-labs/ef-tests/issues/120
- [ ] https://github.com/kkrt-labs/ef-tests/issues/122
- [ ] https://github.com/kkrt-labs/ef-tests/issues/123
- [ ] https://github.com/kkrt-labs/ef-tests/issues/124
- [ ] https://github.com/kkrt-labs/ef-tests/issues/125
- [ ] https://github.com/kkrt-labs/ef-tests/issues/126
```

PR #127, opened 2023-09-20
feat: add stBadOpcode

Eikix:
Time spent on this PR:

Begins work on #89 

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [x] Testing

# What is the new behavior?

-   [ ] Yes
-   [x] No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,8 +30,8 @@ revm-primitives = "1.1"
 reth-rlp = { git = "https://github.com/paradigmxyz/reth.git", tag = "v0.1.0-alpha.7" }
 
 # Kakarot deps
-kakarot-rpc-core = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "0894457" }
-kakarot-test-utils = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "0894457" }
+kakarot-rpc-core = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "ae5e220" }
+kakarot-test-utils = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "ae5e220" }
 
 # Starknet deps
 katana-core = { git = 'https://github.com/dojoengine/dojo', rev = "b924dac" }
--- a/Makefile
+++ b/Makefile
@@ -20,10 +20,10 @@ KAKAROT_COMMIT := .katana/remote_kakarot_sha
 .PHONY: $(EF_TESTS_DIR)
 setup: $(EF_TESTS_DIR)
 
-fetch-dump:
+fetch-dump: fetch-kakarot-submodule-commit
 	cargo run --features dump --bin fetch-dump-katana
 
-$(KAKAROT_COMMIT):
+fetch-kakarot-submodule-commit:
 	cargo run --features fetch-commit --bin fetch-kakarot-submodule-commit
 
 # Runs the Ethereum Foundation tests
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -30,8 +30,8 @@ revm-primitives = "1.1"
 reth-rlp = { git = "https://github.com/paradigmxyz/reth.git", tag = "v0.1.0-alpha.7" }
 
 # Kakarot deps
-kakarot-rpc-core = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "0894457" }
-kakarot-test-utils = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "0894457" }
+kakarot-rpc-core = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "ae5e220" }
```
jobez, line 33: this is the bump for steps right?

crates/ef-testing/src/models/case.rs
```diff
@@ -107,7 +109,14 @@ impl BlockchainTestCase {
...
                 | "OutOfGasContractCreation.json" // ef-test #116
                 | "CallContractToCreateContractAndCallItOOG.json" // ef-test #117
-        )
+                | "undefinedOpcodeFirstByte.json" // ef-tests #121
+                | "measureGas.json" // ef-tests #122
+                | "badOpcodes.json" // ef-tests #123
+                | "operationDiffGas.json" // ef-tests #124
+                | "invalidDiffPlaces.json" // ef-tests #125
+                | "invalidAddr.json" // ef-tests #126
+        ) || matches!(name, name if name.starts_with("opc") && name.ends_with(".json"))
```
jobez, line 118: maybe we can create an issue about having two distinct places to add skips, by specific case or by suite.

jobez (changes requested):

crates/ef-testing/src/models/case.rs
Eikix: wdym

Cargo.toml
Eikix: I don't think so, I think it was a small bump by me when I thought it changed something (it doesn't)

crates/ef-testing/src/models/case.rs
Eikix: ohh skipping a whole suite?

jobez: kk, talked with greg and it seems we arent distinguishing case/suite rn for skips (case is a specific key/val pair within a json file, json file is a suite). what i meant was we'd have a `skipped_by_name` and a `skipped_by_regex` so its less of a scary big list to manage.

Eikix: ah i see

so case: key/val pair within a json
suite: json file
folder of tests: ?

Eikix: you're right! let's create the issue and improve clarity

Makefile
```diff
@@ -20,10 +20,10 @@ KAKAROT_COMMIT := .katana/remote_kakarot_sha
 .PHONY: $(EF_TESTS_DIR)
 setup: $(EF_TESTS_DIR)
 
-fetch-dump:
+fetch-dump: fetch-kakarot-submodule-commit
```
jobez, line 23: this was changed because you had issues with loading the latest commit's dump, oui?

jobez (approved):
left a comment guessing why the makefile changes were included in this, other than that lgtm

Makefile
Eikix: yes, thank you
problem was that if the fetch already existed, it wouldnt fetch once more

PR #144, opened 2023-09-21
feat: invoke sequencer

greged93:
Implements execution for the Kakarot sequencer.

Time spent on this PR: 0.5 day

Resolves: #143 

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
The Kakarot sequencer can now execute simple transactions.

-   [ ] Yes
-   [x] No


Eikix (commented):
can we also have cairo1 programs to run on the test sequencer?

crates/sequencer/src/commit.rs
```diff
@@ -0,0 +1,29 @@
+use blockifier::state::{
+    cached_state::CachedState,
+    state_api::{State as BlockifierState, StateReader as BlockifierStateReader},
+};
+
+pub trait Committer: BlockifierState + BlockifierStateReader {
+    fn commit<S>(cached_state: &mut CachedState<&mut S>)
```
Eikix, line 7: we can pass the generic S to the trait rather, no?

crates/sequencer/src/commit.rs
```diff
@@ -0,0 +1,29 @@
...
+};
+
+pub trait Committer<S>
+where
+    for<'a> &'a mut S: BlockifierState + BlockifierStateReader,
+{
+    fn commit(cached_state: &mut CachedState<&mut S>) {
+        let diff = cached_state.to_state_diff();
+        for (address, class_hash) in diff.address_to_class_hash {
+            let _ = cached_state.state.set_class_hash_at(address, class_hash);
```
Eikix, line 13: why do we need `let _` ? to ignore an error? 
why aren't we handling the error if any?

crates/sequencer/src/commit.rs
```diff
@@ -0,0 +1,29 @@
...
+        for (address, _) in diff.address_to_nonce {
+            let _ = cached_state.state.increment_nonce(address);
+        }
+        for (address, storage_updates) in diff.storage_updates {
+            for (k, v) in storage_updates {
+                cached_state.state.set_storage_at(address, k, v);
+            }
+        }
+        for (class_hash, compiled_class_hash) in diff.class_hash_to_compiled_class_hash {
+            let _ = cached_state
```
Eikix, line 24: same

crates/sequencer/src/commit.rs
```diff
@@ -0,0 +1,29 @@
...
+where
+    for<'a> &'a mut S: BlockifierState + BlockifierStateReader,
+{
+    fn commit(cached_state: &mut CachedState<&mut S>) {
+        let diff = cached_state.to_state_diff();
+        for (address, class_hash) in diff.address_to_class_hash {
+            let _ = cached_state.state.set_class_hash_at(address, class_hash);
+        }
+        for (address, _) in diff.address_to_nonce {
+            let _ = cached_state.state.increment_nonce(address);
```
Eikix, line 16: same

crates/sequencer/src/commit.rs
```diff
@@ -0,0 +1,29 @@
+use blockifier::state::{
+    cached_state::CachedState,
+    state_api::{State as BlockifierState, StateReader as BlockifierStateReader},
+};
+
+pub trait Committer<S>
```
Eikix, line 6: is this naming taken from Blockfier?

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,173 @@ where
...
     }
 }
+
+impl<S> Execution for Sequencer<S>
+where
+    for<'a> &'a mut S: State + StateReader + Committer<S>,
+{
+    fn execute(&mut self, transaction: Transaction) -> Result<(), TransactionExecutionError> {
+        let mut cached_state = CachedState::new(&mut self.state);
+        let res = transaction.execute(&mut cached_state, &self.context, false);
```
Eikix, line 46: can we name the boolean variable? so that reader can see at a glance what is this third argument?

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,173 @@ where
         Self { context, state }
     }
 }
+
+impl<S> Execution for Sequencer<S>
+where
+    for<'a> &'a mut S: State + StateReader + Committer<S>,
+{
+    fn execute(&mut self, transaction: Transaction) -> Result<(), TransactionExecutionError> {
```
Eikix, line 44: why does execute never error out?

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,173 @@ where
...
+
+        match res {
+            Err(err) => {
+                warn!("Transaction execution failed: {:?}", err)
+            }
+            Ok(execution_information) => {
+                <&mut S>::commit(&mut cached_state);
+                match execution_information.revert_error {
+                    Some(err) => {
+                        warn!("Transaction execution failed: {:?}", err)
```
Eikix, line 56:
```suggestion
                        warn!("Transaction execution reverted: {:?}", err)
```

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,173 @@ where
...
+    use crate::constants::test_constants::{
+        FEE_TOKEN_ADDRESS, ONE_BLOCK_NUMBER, ONE_BLOCK_TIMESTAMP, ONE_CLASS_HASH,
+        SEQUENCER_ADDRESS, TEST_ADDRESS, TEST_CONTRACT_ACCOUNT, TEST_CONTRACT_ADDRESS,
+        TWO_CLASS_HASH, ZERO_FELT,
+    };
+    use crate::state::State;
+
+    use super::*;
+
+    fn read_contract_class(path: &str) -> ContractClass {
```
Eikix, line 95:
```suggestion
    fn read_contract_class_v0(path: &str) -> ContractClass {
```

Eikix: could we add some v1 contracts as well?

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,173 @@ where
...
+
+        // When
+        let transaction = Transaction::AccountTransaction(AccountTransaction::Invoke(
+            InvokeTransaction::V1(InvokeTransactionV1 {
+                transaction_hash: TransactionHash(*ZERO_FELT),
+                sender_address: *TEST_CONTRACT_ACCOUNT,
+                calldata: Calldata(
+                    vec![
+                        *TEST_ADDRESS, // destination
+                        FieldElement::from_hex_be(
```
Eikix, line 182: we can use selector macro from starknet-rs

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,173 @@ where
...
+        fund(*TEST_ADDRESS, mutable);
+
+        let context = BlockContext {
+            chain_id: ChainId("KKRT".into()),
+            block_number: *ONE_BLOCK_NUMBER,
+            block_timestamp: *ONE_BLOCK_TIMESTAMP,
+            sequencer_address: *SEQUENCER_ADDRESS,
+            fee_token_address: *FEE_TOKEN_ADDRESS,
+
+            vm_resource_fee_cost: Arc::new(
```
Eikix, line 151: what is the impact of setting n_steps == 1_f64?
why f64?

i understnad its from blockifier but for my culture

crates/sequencer/src/sequencer.rs
greged93: I'm going to check but first I need to compile a cairo 1 program into a json format, not sure if this is possible rn

crates/sequencer/src/commit.rs
greged93: I updated for set_class_hash_at, but increment nonce should never error out.

crates/sequencer/src/sequencer.rs
greged93: I'm not 100% sure of what I am saying, but I think this mapping translates ALL the costs in term of cairo gas. So this maps builtins and steps to a single unit of reference, which I believe in blockifier thel call gas.

Eikix: can we verify and add a comment?

crates/sequencer/src/sequencer.rs
Eikix: ok to do in another issue/pr

crates/sequencer/src/commit.rs
Eikix: I think if it should never run out, it's fine to then handle the error
We never know if one day this'll change in blockifier

crates/sequencer/src/sequencer.rs
greged93: Done!

crates/sequencer/src/commit.rs
greged93: Blockifier uses `commit` and `stage` function names, so I figured Commiter, being an entity that can commit makes sense.

crates/sequencer/src/commit.rs
greged93: fair, will update

crates/sequencer/src/sequencer.rs
```diff
@@ -25,3 +36,240 @@ where
...
+    for<'a> &'a mut S: State + StateReader + Committer<S>,
+{
+    fn execute(&mut self, transaction: Transaction) -> Result<(), TransactionExecutionError> {
+        let mut cached_state = CachedState::new(&mut self.state);
+        let charge_fee = false;
+        let res = transaction.execute(&mut cached_state, &self.context, charge_fee);
+
+        match res {
+            Err(err) => {
+                warn!("Transaction execution failed: {:?}", err);
```
Eikix, line 51: why no bubble up errors?

Eikix (approved):
lgtm

Issue #119, opened 2023-09-20
feat: add the ability to be able to dump and load sequencer state


bajpai244:
# Feature Request

**Describe the Feature Request**

Add the ability to dump state and load a dumped state for the seqeuencer.

**Describe Preferred Solution**

Implement a serializable state, and then add the ability on the seqeuncer to dump and load from the same.

PR #159, opened 2023-09-22
feat: add state dump and loading for the sequencer

bajpai244:
Resolves: #119 

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?

This PR adds the ability to dump and load sequencer state, the serialisable types have been borrowed from dojo's [katana](https://github.com/dojoengine/dojo/tree/main/crates/katana/core/src/db/serde).

-   [ ] Yes
-   [x] No


crates/sequencer/src/serde/contract.rs
```diff
@@ -0,0 +1,246 @@
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use serde::{Deserialize, Serialize};
+use starknet_api::core::EntryPointSelector;
+use starknet_api::deprecated_contract_class::{EntryPoint, EntryPointOffset, EntryPointType};
+
+use super::program::SerializableProgram;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
Eikix, line 10: can you annotate which code is taken from katana? for credit but also to know which code is technically reviewed, tested and works in katana currently

crates/sequencer/src/serde/contract.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn from(value: SerializableEntryPoint) -> Self {
+        Self {
+            selector: value.selector,
+            offset: value.offset.into(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SerializableEntryPointOffset(pub usize);
```
Eikix, line 54: why is this needed?

crates/sequencer/src/serde/contract.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub selector: EntryPointSelector,
+    pub offset: SerializableEntryPointOffset,
+    pub builtins: Vec<String>,
+}
+
+impl From<SerializableEntryPointV1> for blockifier::execution::contract_class::EntryPointV1 {
+    fn from(value: SerializableEntryPointV1) -> Self {
+        blockifier::execution::contract_class::EntryPointV1 {
+            selector: value.selector,
+            offset: value.offset.into(),
```
Eikix, line 79: it looks like the only problem is offset!? that sounds cursed

crates/sequencer/src/state.rs
```diff
@@ -143,6 +170,27 @@ impl BlockifierStateReader for &mut State {
     }
 }
 
+impl From<&SerializableState> for State {
```
Eikix, line 173: Why are we implementing the trait on a reference?

crates/sequencer/src/state.rs
```diff
@@ -143,6 +170,27 @@ impl BlockifierStateReader for &mut State {
...
+    fn from(serializable_state: &SerializableState) -> Self {
+        let mut state = Self {
+            classes: FxHashMap::default(),
+            compiled_class_hashes: serializable_state.compiled_classes_hash.clone(),
+            contracts: serializable_state.contracts.clone(),
+            storage: serializable_state.storage.clone(),
+            nonces: serializable_state.nonces.clone(),
+        };
+
+        serializable_state.classes.iter().for_each(|(class_hash, contract_class)|{
```
Eikix, line 183: why do we need to do this?

crates/sequencer/src/state.rs
```diff
@@ -143,6 +170,27 @@ impl BlockifierStateReader for &mut State {
...
+           state.classes.insert(
+               *class_hash,
+               contract_class.class.clone().try_into().unwrap_or_else(|error| panic!("failed to convert SerializableClassRecord to ContractClass for class_hash {},\n error {}", class_hash, error))
```
Eikix, line 186: must we panic here?
can't we implement TryFrom and return a result?

crates/sequencer/src/state.rs
```diff
@@ -253,4 +312,75 @@ mod tests {
...
+    #[test]
+    pub fn dump_and_load_state() {
+        let mut state = State::default();
+
+        // setting up entry for state.classes
+        let class_hash = ClassHash(stark_felt!("0x1"));
+        let contract_class = get_contract_class(include_str!(
+            "./test_data/contracts/compiled/universal_deployer.json"
+        ));
+        let compiled_class_hash = CompiledClassHash(stark_felt!("0x1"));
```
Eikix, line 325: can you use @greged93 's constants? I think he wrote a similar flow

Eikix: in the constant.rs file

crates/sequencer/src/state.rs
```diff
@@ -22,11 +27,33 @@ use starknet_api::{
...
+            reason: format!("failed to write dump state to the path {:?}", path),
+            context: error,
+        })?;
+
+        Ok(())
+    }
+
+    /// This will read a dump from a file and initialize the state from it
+    pub fn load_state_from_file(path: &PathBuf) -> Result<Self, SerializationError> {
+        let serilizable_state = SerializableState::load_state(path)?;
```
Eikix, line 53:
```suggestion
        let serializable_state = SerializableState::load_state(path)?;
```

crates/sequencer/src/state.rs
```diff
@@ -22,11 +27,33 @@ use starknet_api::{
...
+
+        Ok(Self::from(&serilizable_state))
```
Eikix, line 55:
```suggestion
        Ok(Self::from(&serializable_state))
```

crates/sequencer/src/serde/utils.rs
```diff
@@ -0,0 +1,37 @@
...
+) -> Result<(FieldElement, ContractClass)> {
+    let class_hash = contract_class.class_hash();
+    let contract_class = rpc_to_cairo_contract_class(contract_class)?;
+    let casm_contract = CasmContractClass::from_contract_class(contract_class, true)?;
+    Ok((class_hash, ContractClass::V1(casm_contract.try_into()?)))
+}
+
+/// Converts `starknet-rs` RPC [FlattenedSierraClass] type to Cairo's
+/// [ContractClass](cairo_lang_starknet::contract_class::ContractClass) type.
+pub fn rpc_to_cairo_contract_class(
```
Eikix, line 17: why do we need this?

crates/sequencer/src/serde/utils.rs
```diff
@@ -0,0 +1,37 @@
...
+        entry_points_by_type: serde_json::from_value(value["entry_points_by_type"].clone())?,
+        contract_class_version: serde_json::from_value(value["contract_class_version"].clone())?,
+        sierra_program_debug_info: serde_json::from_value(
+            value["sierra_program_debug_info"].clone(),
+        )
+        .ok(),
+    })
+}
+
+pub fn get_contract_class(contract_class_str: &str) -> ContractClass {
```
Eikix, line 34: this doesn't look like a getter but rather an `into_contract_class`

crates/sequencer/src/serde/state.rs
```diff
@@ -0,0 +1,156 @@
...
+};
+use thiserror::Error;
+
+use crate::state::State;
+
+use super::contract::SerializableContractClass;
+
+/// TODO: wrap the underlying errors in them
+#[derive(Error, Debug)]
+pub enum SerializationError {
```
greged93, line 18: I think we can rework this a little to use more the syntax from thiserror examples?
```
#[derive(Error, Debug)]
pub enum MyError {
    Io {
        #[from]
        source: io::Error,
        backtrace: Backtrace,
    },
}
```
[source](https://docs.rs/thiserror/latest/thiserror/#details)

crates/sequencer/src/serde/state.rs
```diff
@@ -0,0 +1,156 @@
...
+    #[error("{reason:?}")]
+    SerdeJsonError {
+        reason: String,
+        context: serde_json::Error,
+    },
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize, Default)]
+pub struct SerializableState {
+    pub classes: FxHashMap<ClassHash, SerializableClassRecord>,
```
greged93, line 33: Any reason to have a `SerializableClassRecord` instead of just a `SerializableContractClass`?

crates/sequencer/src/serde/state.rs
```diff
@@ -0,0 +1,156 @@
...
+    ) -> Result<FxHashMap<ContractStorageKey, StarkFelt>, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let transformed = HashMap::<String, String>::deserialize(deserializer)?;
+        let mut map = FxHashMap::default();
+        for (key_str, value_str) in transformed.iter() {
+            let contract_storage_key: ContractStorageKey = serde_json::from_str(key_str)
+                .unwrap_or_else(|error| {
+                    panic!(
```
greged93, line 138: Could we return an error instead of panicking?

crates/sequencer/src/serde/state.rs
```diff
@@ -0,0 +1,156 @@
...
+                        "failed to deserialize contract_storage_key {},\n error {}",
+                        key_str, error
+                    )
+                });
+
+            let storage_value: StarkFelt =
+                serde_json::from_str(value_str).unwrap_or_else(|error| {
+                    panic!(
```
greged93, line 146: Same, should avoid panicking

crates/sequencer/src/serde/utils.rs
greged93: Also we should avoid `get` prefix

crates/sequencer/src/serde/state.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub classes: FxHashMap<ClassHash, SerializableClassRecord>,
+    pub compiled_classes_hash: FxHashMap<ClassHash, CompiledClassHash>,
+    pub contracts: FxHashMap<ContractAddress, ClassHash>,
+    #[serde(with = "serialize_contract_storage")]
+    pub storage: FxHashMap<ContractStorageKey, StarkFelt>,
+    pub nonces: FxHashMap<ContractAddress, Nonce>,
+}
+
+impl SerializableState {
+    pub fn dump_state(&self) -> Result<Vec<u8>, SerializationError> {
```
greged93, line 42: I'm wondering: should we rename this? This converts the `SerializableState` into a Vec<u8>, so naming it `dump_state` sounds too specific.

crates/sequencer/src/serde/state.rs
```diff
@@ -0,0 +1,156 @@
...
+        let serialized_state_buffer =
+            serde_json::to_vec(&self).map_err(|error| SerializationError::SerdeJsonError {
+                reason: "Failed at serializing state to a buffer".to_string(),
+                context: error,
+            })?;
+
+        Ok(serialized_state_buffer)
+    }
+
+    pub fn load_state(path: &PathBuf) -> Result<Self, SerializationError> {
```
greged93, line 52: could we change this to `new`? `new` could take an Option<PathBuf> and load the file if Option is not None.

crates/sequencer/src/state.rs
greged93: I think we can, I have almost all values in there, and we can add if missing.

crates/sequencer/src/state.rs
```diff
@@ -253,4 +312,75 @@ mod tests {
...
+            loaded_state.contracts.get(&contract_address)
+        );
+        assert_eq!(
+            state.storage.get(&contract_storage_key),
+            loaded_state.storage.get(&contract_storage_key)
+        );
+        assert_eq!(
+            state.nonces.get(&contract_address),
+            loaded_state.nonces.get(&contract_address)
+        );
```
greged93, line 377: Can't we derive `PartialEq` and just have one `assert_eq!`?

crates/sequencer/src/test_data/contracts/compiled/cairo1_contract.json
```diff
@@ -0,0 +1,2534 @@
+{
```
greged93, line 1: You can remove both these contracts and use the ones I have added.

greged93 (changes requested):
Couple of comments, great work!

crates/sequencer/src/serde/contract.rs
bajpai244: Makes sense, completed via [this](https://github.com/kkrt-labs/ef-tests/pull/159/commits/d7001aacb5a993ccb1621910a171bd4efb25b9ef) commit.

crates/sequencer/src/serde/contract.rs
bajpai244: This has been taken from Katana directly.

crates/sequencer/src/serde/contract.rs
bajpai244: This has been taken from Katana directly.

crates/sequencer/src/state.rs
bajpai244: So that we want to create a `State` from `SerializableState`, we can do so without moving the value and rather than borrowing it.

crates/sequencer/src/state.rs
bajpai244: The `contract_class` is of type `SerializableClassRecord`, hence we need a conversion from this to `ContractClass`  
{ which is not seriaizable, and hence we have a seriaizable type for it } which is what we store in a normal state of the seqeuncer.

image

crates/sequencer/src/state.rs
bajpai244: I think panic here is fine right? Because this situation is a bug, it means that there is an error in our serialisation for contract class { which is already tested by dojo! }, and from my understanding, panics are reserved for bugs, hence, from our assumption, this should never happen!

crates/sequencer/src/state.rs
bajpai244: makes sense, done via [this](https://github.com/kkrt-labs/ef-tests/pull/159/commits/3c67a7a263c39f7e900dc289da5db5f2f365fb4a) commit.

crates/sequencer/src/state.rs
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/159/commits/536b155d1182f5c3353afda2d253ecdf3e71582a) commit.

crates/sequencer/src/state.rs
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/159/commits/536b155d1182f5c3353afda2d253ecdf3e71582a) commit.

crates/sequencer/src/serde/utils.rs
bajpai244: required in the serialization of contract class, required from a dojo borrowed file for contract serialization.

image

crates/sequencer/src/serde/utils.rs
bajpai244: This has been borrowed directly from katana and it is also used from the dojo borrowed file.

bajpai244: I have made a [commit](https://github.com/kkrt-labs/ef-tests/pull/159/commits/194d691759b7eb2644b438bb8bde051f260a4994) resolving it.

crates/sequencer/Cargo.toml
```diff
@@ -12,14 +12,21 @@ license.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+anyhow = "1.0.75"
 # Starknet
 # TODO: remove the blockifier patch on the workspace once we can remove Katana.
 blockifier = { package = "blockifier", git = "https://github.com/starkware-libs/blockifier.git", tag = "v0.3.0-rc0" }
+cairo-lang-starknet = { workspace = true }
+cairo-vm = { workspace = true }
+serde.workspace = true
```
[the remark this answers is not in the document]
bajpai244: resolved [here](https://github.com/kkrt-labs/ef-tests/pull/159/commits/bd6c71691c5b4e0f5e457503cb27640f93f078ae).

crates/sequencer/src/state.rs
bajpai244: resolved [here](https://github.com/kkrt-labs/ef-tests/pull/159/commits/bd6c71691c5b4e0f5e457503cb27640f93f078ae).

crates/sequencer/src/serde/contract.rs
```diff
@@ -0,0 +1,246 @@
...
+            builtins: value.builtins,
+        }
+    }
+}
+
+impl TryFrom<SerializableContractClass> for blockifier::execution::contract_class::ContractClass {
+    type Error = anyhow::Error;
+
+    fn try_from(value: SerializableContractClass) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
bajpai244: @greged93 this whole file is borrowed from Dojo, they said they are working on serialization for these types, if we still need this file in sometime, then might be we can refactor this otherwise we will probably not need these serde types in sometime!

crates/sequencer/src/serde/contract.rs
greged93: but can it be changed?

crates/sequencer/src/state.rs
greged93: I don't really see the need for this, specially as the implementation itself just copies everything? So in the end we are copying values

crates/sequencer/src/state.rs
greged93: I think we should never have a `panic!` in our code, would make a lot more sense to rely the error upstream if we have one

crates/sequencer/Cargo.toml
```diff
@@ -12,14 +12,21 @@ license.workspace = true
...
 starknet_api = { workspace = true }
 starknet = { workspace = true }
+thiserror.workspace = true
 
 # Other
 rustc-hash = "1.1.0"
 
 [dev-dependencies]
 lazy_static = { workspace = true }
+starknet = { workspace = true }
```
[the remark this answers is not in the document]
bajpai244: resolved [here](https://github.com/kkrt-labs/ef-tests/pull/159/commits/bd6c71691c5b4e0f5e457503cb27640f93f078ae).

crates/sequencer/src/serde/contract.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use blockifier::execution::contract_class::ContractClass;
+    use starknet::core::types::contract::SierraClass;
+
+    use super::*;
+    use crate::serde::utils::{get_contract_class, rpc_to_inner_class};
```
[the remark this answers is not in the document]
bajpai244: required by dojo tests, I don't think we should  change this, in sometime we should not need it mostly as discussed once we have serialization enabled for ContractClass.

crates/sequencer/src/serde/state.rs
bajpai244: borrowed this name, but can change this.

bajpai244: I just realized this type already exists, and we can remove SerializableClassRecord  as it simply wraps it.

bajpai244: fixed [here](https://github.com/kkrt-labs/ef-tests/pull/159/commits/aef6614acb2835dcfef73d0917412906a9e8350c).

crates/sequencer/src/serde/contract.rs
bajpai244: This is required for the serialization.

crates/sequencer/src/state.rs
bajpai244: But if you will do it directly on type `SerializableState`, then this copying will happen twice right?

Firstly you have to call this method on a clone, and then there will another copying inside the implementation. I think the reference approach is hence more optimizied.

Lemme know what you think.

crates/sequencer/src/serde/state.rs
bajpai244: how about `serialize_state`?

crates/sequencer/src/serde/state.rs
bajpai244: I think, `from_file_path` is a better approach, it should be a static method and do one single thing!

We can see this pattern in other rust codebases as well, where there is options like `from_string`, etc.

bajpai244: Lemme know what you think about it.

crates/sequencer/src/state.rs
bajpai244: Great suggestion! Implemented via [this](https://github.com/kkrt-labs/ef-tests/pull/159/commits/87e3c996cdad92f3242f915f5666c559fb74b374) commit.

greged93:
We will fork cairo-vm and blockifier in order to implement serialize on the necessary types.

PR #316, opened 2023-09-27
feat: evm sequencer

greged93:
Provides a wrapper for a evm sequencer around the kakarot sequencer.

Time spent on this PR: 2 days

Resolves: #267 

# Pull Request type

-   [ ] Bugfix
-   [x] Feature
-   [ ] Build-related changes
-   [ ] Testing

# What is the new behavior?
Provides an easy way to completely deploy contracts, eoa, kakarot and fund addresses.

-   [ ] Yes
-   [x] No


.github/workflows/test.yml
```diff
@@ -1,16 +1,20 @@
...
 
 on:
-  push:
-    branches: [main]
-  pull_request:
+  workflow_call:
 
 jobs:
   test:
     runs-on: ubuntu-latest-16-cores
```
ClementWalter, line 8: do we really need this expensive runner?

.github/workflows/test.yml
ClementWalter, line 1: you can remove the commented lines as well plz

      # fetch ef tests
      # fetch dump
      # run tests

Cargo.toml
```diff
@@ -34,6 +34,8 @@ kakarot-rpc-core = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev =
 kakarot-test-utils = { git = "https://github.com/kkrt-labs/kakarot-rpc.git", rev = "ae5e220" }
 
 # Starknet deps
+cairo-vm = "0.8.2"
+blockifier = { package = "blockifier", git = "https://github.com/starkware-libs/blockifier.git", tag = "v0.3.0-rc0" }
```
ClementWalter, line 38: is pinning the tag enough? tag can be overridden

crates/sequencer/src/transaction.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl TryFrom<StarknetTransaction> for ExecutionTransaction {
+    type Error = eyre::Error;
+
+    fn try_from(transaction: StarknetTransaction) -> Result<Self, Self::Error> {
+        match transaction.0 {
+            BroadcastedTransaction::Invoke(invoke) => Ok(ExecutionTransaction::AccountTransaction(
+                AccountTransaction::Invoke(InvokeTransaction::V1(InvokeTransactionV1 {
+                    transaction_hash: TransactionHash(Into::<StarkHash>::into(
+                        Into::<StarkFelt>::into(FieldElement::ONE), // TODO: Replace with actual computed hash.
```
ClementWalter, line 32: is there a risk to have a weird bug somewhere because of this? it is difficult to put the real hash?

crates/ef-testing/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 pub mod models;
+pub mod sequencer;
```
ClementWalter, line 2: there is a sequencer crate, imported in the ef-testing crate, but the ef-testing crate also has a sequencer module? this is a bit confusing

crates/ef-testing/src/sequencer/constants.rs
```diff
@@ -0,0 +1,101 @@
...
+use starknet::core::types::{contract::legacy::LegacyContractClass, FieldElement};
+use starknet_api::{
+    block::{BlockNumber, BlockTimestamp},
+    core::{ChainId, ClassHash, ContractAddress, PatriciaKey},
+    hash::StarkFelt,
+};
+
+lazy_static! {
+    // Chain params
+    pub static ref CHAIN_ID: u64 = 1_263_227_476;
```
ClementWalter, line 14: I'd rather have 0x4b4b5254 literal `0x4b4b5254` so that it's clearer that it's the ascii value of some string

crates/ef-testing/src/sequencer/constants.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // Vm resources
+    pub static ref VM_RESOURCES: HashMap<String, f64> = [
+        (String::from("n_steps"), 1f64),
```
ClementWalter, line 18: can you replace `1f64` with `1_f64`?

crates/ef-testing/src/sequencer/constants.rs
```diff
@@ -0,0 +1,101 @@
...
+    core::{ChainId, ClassHash, ContractAddress, PatriciaKey},
+    hash::StarkFelt,
+};
+
+lazy_static! {
+    // Chain params
+    pub static ref CHAIN_ID: u64 = 1_263_227_476;
+
+    // Vm resources
+    pub static ref VM_RESOURCES: HashMap<String, f64> = [
```
ClementWalter, line 17: is this the cost of each builtin?

crates/ef-testing/src/sequencer/constants.rs
```diff
@@ -0,0 +1,101 @@
...
+        ("ec_op_builtin".to_string(), 1f64),
+        ("keccak_builtin".to_string(), 1f64),
+        ("segment_arena_builtin".to_string(), 1f64),
+    ]
+    .into_iter()
+    .collect();
+
+    // Block context
+    pub static ref BLOCK_CONTEXT: BlockContext = BlockContext {
+            chain_id: ChainId("KKRT".into()),
```
ClementWalter, line 34: ❌ no duplicated definition of chain_id plz

crates/ef-testing/src/sequencer/constants.rs
```diff
@@ -0,0 +1,101 @@
...
+            block_number: BlockNumber(0),
+            block_timestamp: BlockTimestamp(0),
+            sequencer_address: *SEQUENCER_ADDRESS,
+            fee_token_address: *FEE_TOKEN_ADDRESS,
+            vm_resource_fee_cost: Arc::new(VM_RESOURCES.clone()),
+            gas_price: 1,
+            invoke_tx_max_n_steps: 2u32.pow(24),
```
ClementWalter, line 41: can we always use `_` as a sep between number and type?

crates/ef-testing/src/sequencer/constants.rs
```diff
@@ -0,0 +1,101 @@
...
+    use revm_primitives::B256;
+
+    lazy_static! {
+        pub static ref PRIVATE_KEY: B256 =
+            B256::from_str("0x6ae82d865482a203603ecbf25c865e082396d7705a6bbce92c1ff1d6ab9b503c")
+                .unwrap();
+        pub static ref PUBLIC_KEY: Address =
+            Address::from_str("0x7513A12F74fFF533ee12F20EE524e4883CBd1945").unwrap();
+        pub static ref TEST_CONTRACT_ADDRESS: Address = Address::from_low_u64_be(10);
+        pub static ref SELECTOR: Bytes = Bytes::from_str("0x693c6139").unwrap();
```
ClementWalter, line 99: what is this selector? I was not able to fund in 4byte.directory

crates/ef-testing/src/sequencer/types.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+impl From<FeltSequencer> for StarkFelt {
+    fn from(felt: FeltSequencer) -> Self {
+        StarkFelt::from(felt.0)
+    }
+}
+
+impl TryFrom<FeltSequencer> for ContractAddress {
```
ClementWalter, line 37: how could a Felt not fit into a contract address?

crates/ef-testing/src/sequencer/utils.rs
```diff
@@ -0,0 +1,92 @@
...
+};
+use starknet_api::{
+    core::{ClassHash, ContractAddress},
+    hash::StarkFelt,
+    state::StorageKey,
+};
+
+use super::{constants::KAKAROT_ADDRESS, KakarotSequencer};
+
+pub(crate) fn get_storage_var_address(
```
ClementWalter, line 17: do we really need this wrapper?

crates/ef-testing/src/sequencer/utils.rs
```diff
@@ -0,0 +1,92 @@
...
+}
+
+pub(crate) fn split_u256(value: U256) -> [u128; 2] {
+    [
+        (value & U256::from(u128::MAX)).try_into().unwrap(), // safe unwrap <= U128::MAX.
+        (value >> 128).try_into().unwrap(),                  // safe unwrap <= U128::MAX.
+    ]
+}
+
+pub(crate) fn contract_address_to_starkfelt(contract_address: &ContractAddress) -> StarkFelt {
```
ClementWalter, line 42: in the `types.rs` you defined a very similar `TryFrom<FeltSequencer> for ContractAddress`, + I'm wondering why we should do this since these two types are from starknet_api

crates/ef-testing/src/sequencer/utils.rs
```diff
@@ -0,0 +1,92 @@
...
+    *contract_address.0.key()
+}
+
+pub(crate) fn class_hash_to_starkfelt(class_hash: &ClassHash) -> StarkFelt {
```
ClementWalter, line 46: same here, I'm wondering why we should do this since these two types are from starknet_api

crates/ef-testing/src/sequencer/utils.rs
```diff
@@ -0,0 +1,92 @@
...
+    class_hash.0
+}
+
+pub fn bytes_to_felt_vec(bytes: &Bytes) -> Vec<FieldElement> {
```
ClementWalter, line 50: looks very similar to `split_bytecode_to_starkfelt(bytecode: &Bytes) -> Vec<StarkFelt>`, do we need both?

crates/ef-testing/src/sequencer/utils.rs
```diff
@@ -0,0 +1,92 @@
...
+    bytes.to_vec().into_iter().map(FieldElement::from).collect()
+}
+
+#[allow(dead_code)]
+pub(crate) fn to_broadcasted_starknet_transaction(
```
ClementWalter, line 55: what is this used for? also, even though rust compiler resolves circular deps, I'm always quite worried about their introduction as it often is a sign a "feature envy"

crates/ef-testing/src/sequencer/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+            &CONTRACT_ACCOUNT_CLASS_HASH,
+            ContractClass::V0(ContractClassV0::try_from_json_string(
+                &serde_json::to_string(&*CONTRACT_ACCOUNT_CLASS)
+                    .map_err(|err| StateError::ProgramError(ProgramError::Parse(err)))?,
+            )?),
+        )?;
+        (&mut self.0.state).set_contract_class(
+            &EOA_CLASS_HASH,
+            ContractClass::V0(ContractClassV0::try_from_json_string(
+                &serde_json::to_string(&*EOA_CLASS)
```
ClementWalter, line 101: same about serde

crates/ef-testing/src/sequencer/utils.rs
ClementWalter: and actuallt ("suprisingly") the `Kakarot.compute_starknet_address` doesn't even use `self`

crates/ef-testing/src/sequencer/setup.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+impl InitializeEvmState for KakarotSequencer {
+    fn initialize_contract(
+        &mut self,
+        evm_address: &Address,
+        bytecode: &Bytes,
+        nonce: U256,
+        evm_storage: Vec<(U256, U256)>,
+    ) -> StateResult<()> {
```
ClementWalter, line 41: why pushing heaps of times and not only once?

crates/ef-testing/src/sequencer/setup.rs
```diff
@@ -0,0 +1,263 @@
...
+    split_bytecode_to_starkfelt, split_u256,
+};
+use super::KakarotSequencer;
+
+pub trait InitializeEvmState {
+    fn initialize_contract(
+        &mut self,
+        evm_address: &Address,
+        bytecode: &Bytes,
+        nonce: U256,
```
ClementWalter, line 27: is the nonce really a `U256`? I think that geth/reth use u64

crates/ef-testing/src/sequencer/setup.rs
```diff
@@ -0,0 +1,263 @@
...
+            }
+        })?);
+        let nonce_storage = (
+            get_storage_var_address("nonce", &[]).unwrap(), // safe unwrap: var is ASCII
+            nonce,
+        );
+        storage.push(nonce_storage);
+
+        // Initialize the storage vars.
+        let is_evm_storage_empty = evm_storage.is_empty();
```
ClementWalter, line 105: plz define var closer the usage

ClementWalter: is actually createing a variable necesary?

crates/ef-testing/src/sequencer/setup.rs
```diff
@@ -0,0 +1,263 @@
...
+        storage.push(kakarot_address_storage);
+
+        // Initialize the nonce storage var.
+        let nonce = StarkFelt::from(TryInto::<u128>::try_into(nonce).map_err(|err| {
+            StarknetApiError::OutOfRange {
+                string: err.to_string(),
+            }
+        })?);
+        let nonce_storage = (
+            get_storage_var_address("nonce", &[]).unwrap(), // safe unwrap: var is ASCII
```
ClementWalter, line 99: nonce management depends on the EOA/CA type

crates/ef-testing/src/sequencer/setup.rs
```diff
@@ -0,0 +1,263 @@
...
+        storage.push(implementation_storage);
+
+        // Write all the storage vars to the sequencer state.
+        let starknet_address = starknet_address.try_into()?;
+        for (k, v) in storage {
+            (&mut self.0.state).set_storage_at(starknet_address, k, v);
+        }
+
+        // Set up the contract class hash and nonce.
+        self.0.state.set_nonce(starknet_address, Nonce(nonce));
```
ClementWalter, line 136: so you indeed set up nonce both in storage and in starknet state to avoid making an if/else?

.github/workflows/test.yml
greged93: Not for now, my mistake

crates/sequencer/src/transaction.rs
greged93: I don't think we ever use the transaction hash, and it is [not complicated](https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#v1_hash_calculation) but just adds some overhead which I don't think is very useful right now

crates/ef-testing/src/lib.rs
greged93: want me to rename to `evm-sequencer`?

crates/ef-testing/src/sequencer/constants.rs
greged93: Of each builtin and of steps.

crates/ef-testing/src/sequencer/constants.rs
greged93: It was actually not used, came from ethereum/tests repo but wasn't necessary

crates/ef-testing/src/sequencer/types.rs
greged93: Contract address is a wrapper around PatriciKey, and PatriciaKey upper bound is 2^251, lower than the felt limit

crates/ef-testing/src/sequencer/utils.rs
greged93: Just easier for error management for now, allows me to not start to worry about the error I return

crates/ef-testing/src/sequencer/utils.rs
greged93: Sometimes I need to get the Felt inside the contract address, I can remove this and replace everywhere with `contract_address.0.key()`, just find it nicer like this

crates/ef-testing/src/sequencer/utils.rs
greged93: same as above, can remove and replace everwhere by `class_hash.0`, just thought it was more explicit and easier to read.

crates/ef-testing/src/sequencer/utils.rs
greged93: it's actually quite different: `bytes_to_felt_vec` will split the Bytes object into `u8` and convert each of them to a FieldElement. `split_bytecode_to_starkfelt` will split the bytecode into chunks of 16 bytes in order to load them into the contract storage in the correct way.

crates/ef-testing/src/sequencer/utils.rs
greged93: removed the self which wasn't needed indeed. I am not sure what you mean by circular dep in this case. This function is used to convert a rlp encoded evm transaction into a starknet transaction.

crates/ef-testing/src/sequencer/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+            (
+                get_storage_var_address("externally_owned_account_class_hash", &[]).unwrap(),
+                class_hash_to_starkfelt(&EOA_CLASS_HASH),
+            ),
+            (
+                get_storage_var_address("account_proxy_class_hash", &[]).unwrap(),
+                class_hash_to_starkfelt(&PROXY_CLASS_HASH),
+            ),
+        ];
+        storage.append(kakarot_class_hashes);
```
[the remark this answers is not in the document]
greged93: I could do that, just splitting in this way makes it more readable and doesn't add much overhead imo

crates/ef-testing/src/sequencer/mod.rs
```diff
@@ -0,0 +1,137 @@
...
+        for (k, v) in storage {
+            (&mut self.0.state).set_storage_at(*KAKAROT_ADDRESS, k, v);
+        }
+
+        // Write the kakarot class and class hash.
+        (&mut self.0.state).set_class_hash_at(*KAKAROT_ADDRESS, *KAKAROT_CLASS_HASH)?;
+        (&mut self.0.state).set_contract_class(
+            &KAKAROT_CLASS_HASH,
+            ContractClass::V0(ContractClassV0::try_from_json_string(
+                &serde_json::to_string(&*KAKAROT_CLASS)
```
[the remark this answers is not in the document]
greged93: Issue is that `KAKAROT_CLASS` is a type from starknet rs, and is useful because it has a function `class_hash()` which allows to compute the class hash of the class easily. However, this function is not present on the blockifier contract class type, that is needed to set the contract class in storage. 

I could implement a conversion function, but given the fact that this setup should only be used to make a dump and is already quite fast (~ 10 seconds), I did not see the need for this.

crates/ef-testing/src/sequencer/setup.rs
greged93: To make the code more readable

crates/ef-testing/src/sequencer/setup.rs
greged93: The nonce we get in the ef-tests is a `JsonU256`type. I can make the sequencer use U64 if you prefer.

crates/ef-testing/src/sequencer/setup.rs
greged93: I was using `into_iter` instead of `iter`, updated

crates/ef-testing/src/sequencer/setup.rs
greged93: Yes, figured it wasn't a huge overhead. Actually the whole `initialize_contract` function is account agnostic and works for both accounts type, only if/else needed was for the `_implementation` storage var. Figured this was easier over having different setups per account type.

Cargo.toml
greged93: I just pinned to the latest tag, I don't think they would override with a breaking change would they? I can pin to a commit if you want

crates/sequencer/src/transaction.rs
ClementWalter: I'm just scared that we forgot this and one day face a bugs. I prefer if we put the real hash

crates/ef-testing/src/sequencer/utils.rs
ClementWalter: but you're just wrapping the blockifier with no change?

crates/ef-testing/src/sequencer/utils.rs
ClementWalter: `sequencer/utils.rs` imports `use super::{constants::KAKAROT_ADDRESS, KakarotSequencer};`

and `mod.rs` imports

```
use self::utils::{
    class_hash_to_starkfelt, contract_address_to_starkfelt, get_storage_var_address,
};

pub(crate) struct KakarotSequencer(Sequencer<State>);
```

so A imports B imports A, it's circular deps and it wouldn't work in python for example.

It means here that the `compute_starknet_address` should probably not be in the `KakarotSequencer`

crates/ef-testing/src/sequencer/mod.rs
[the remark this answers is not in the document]
ClementWalter: it adds the overhead to understand why it's split. Especially here it's just the same `get_storage_var.., class_hash_to...`, and you mix `vec` and unit `push`, this is just confusing

crates/ef-testing/src/sequencer/setup.rs
ClementWalter: just like inlining a loop is more readable?

crates/ef-testing/src/sequencer/setup.rs
ClementWalter: you do one if/else, and I'm not really in favor of creating a setup that is not 100% identical to what it's supposed to be

crates/ef-testing/src/sequencer/utils.rs
greged93: Yeah just wrapping and returning a eyre::Error

crates/ef-testing/src/sequencer/utils.rs
greged93: Could implement `From` for it yes

crates/ef-testing/src/sequencer/utils.rs
greged93: will move it to utils

crates/ef-testing/src/sequencer/setup.rs
greged93: I am using u256 because this is the type used byt reth in the [ef-tests for the nonce](https://github.com/paradigmxyz/reth/blob/main/testing/ef-tests/src/models.rs#L208). I can make it u64, just means we will have a try_from somewhere else in the code.

crates/ef-testing/src/sequencer/setup.rs
greged93: so you prefer one big if else based on the account then?

crates/ef-testing/src/sequencer/utils.rs
greged93: I'm worried the PR will take a lot of time to get merged. I can make one but last merged PR on starknet api was 3 weeks ago

crates/ef-testing/src/sequencer/setup.rs
greged93: not sure what you mean, but you want me to change this then?

crates/ef-testing/src/evm_sequencer/setup.rs
```diff
@@ -0,0 +1,246 @@
...
+    PROXY_CLASS_HASH,
+};
+use super::types::FeltSequencer;
+use super::utils::{
+    class_hash_to_starkfelt, compute_starknet_address, contract_address_to_starkfelt,
+    split_bytecode_to_starkfelt, split_u256,
+};
+use super::KakarotSequencer;
+
+pub trait InitializeEvmState {
```
ClementWalter, line 22: should this be named instead something like `EvmState`, and consequently

```
pub trait EvmState {
    fn setup_account(
        &mut self,
        address // no need for evm_ here
// ..
    fn fund_account(
```

crates/ef-testing/src/evm_sequencer/setup.rs
```diff
@@ -0,0 +1,246 @@
...
+            StarknetApiError::OutOfRange {
+                string: err.to_string(),
+            }
+        })?);
+        let starknet_address = compute_starknet_address(evm_address);
+        let evm_address = Into::<FeltSequencer>::into(*evm_address).into();
+
+        let mut storage = vec![
+            (
+                get_storage_var_address("evm_address", &[]).unwrap(), // safe unwrap: var is ASCII
```
ClementWalter, line 52: can't we just use `?` and possibly update the signature of `initialize` to return `Result`?
This wouldn't shock me to have this function possibly failing, if for example one tries to initialize two times the same address

crates/ef-testing/src/evm_sequencer/setup.rs
```diff
@@ -0,0 +1,246 @@
...
+    ) -> StateResult<()> {
+        let nonce = StarkFelt::from(TryInto::<u128>::try_into(nonce).map_err(|err| {
+            StarknetApiError::OutOfRange {
+                string: err.to_string(),
+            }
+        })?);
+        let starknet_address = compute_starknet_address(evm_address);
+        let evm_address = Into::<FeltSequencer>::into(*evm_address).into();
+
+        let mut storage = vec![
```
ClementWalter, line 50: can't we do the call to `get_storage_var_address` in the loop and not inlining it here for all occurences?

crates/ef-testing/src/evm_sequencer/setup.rs
```diff
@@ -0,0 +1,246 @@
...
+                get_storage_var_address("kakarot_address", &[]).unwrap(), // safe unwrap: var is ASCII
+                contract_address_to_starkfelt(&KAKAROT_ADDRESS),
+            ),
+            (
+                get_storage_var_address("nonce", &[]).unwrap(), // safe unwrap: var is ASCII
+                nonce,
+            ),
+            (
+                get_storage_var_address("_implementation", &[]).unwrap(), // safe unwrap: var is ASCII
+                if bytecode.is_empty() && evm_storage.is_empty() {
```
ClementWalter, line 77: here I would set the nonce i the if/else as well. There are all the shared things, and one if/else for everything EOA vs CA

crates/sequencer/src/transaction.rs
```diff
@@ -0,0 +1,93 @@
...
+                    )),
+                })),
+            )),
+            // TODO: Add support for other transaction types.
+            _ => Err(eyre::eyre!("Unsupported transaction type")),
+        }
+    }
+}
+
+const PREFIX_INVOKE: FieldElement = FieldElement::from_mont([
```
ClementWalter, line 69: while this is (probably) true, keeping here a hex string and doing would be more explicit

https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transactions/#v1_hash_calculation

ClementWalter (changes requested):
minors things

crates/ef-testing/src/evm_sequencer/setup.rs
greged93: `get_storage_var_address` just returns the address of the storage var so it shouldn't fail, actually since we are still on an old commit of blockifier it still returns an error but this has been updated in a recent PR: https://github.com/starkware-libs/blockifier/commit/0bdd300140b21fd0a8accb51621341d7e668a5a9

crates/ef-testing/src/evm_sequencer/setup.rs
greged93: I'm not sure what you mean, you want something like this?
```
storage_vars = ["evm_address, "is_initialized_", ...];
let storage = storage_vars.into_iter().map(|x| get_storage_var_address(x, &[]).unwrap()).collect();
```

crates/ef-testing/src/evm_sequencer/setup.rs
greged93: there are quite a lot of things that aren't shared, the reason I did it that way is that setting one variable in the storage doesn't change much imo and makes the code easier to read.
The differences between eoa and ca are:
- eoa: kakarot_address
- ca: bytecode, bytecode_len_, storage_, nonce 

The similarities are: evm_address, is_initialized_, _implementation
To note is that since bytecode and storage will be empty for eoa, the only values being set that aren't needed are nonce and bytecode_len_

crates/sequencer/src/transaction.rs
greged93: this is copied from starknet-rs but I can udpate

ClementWalter (approved):

PR #466, opened 2023-10-03
Increase clipy timeout in trunk

ClementWalter:


Eikix:
Unclear why the CI is still failing

PR #468, opened 2023-10-04
fix: kakarot sequencer

greged93:
# Pull Request type

Resolves: #NA

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?
Fee token is deployed in the initialization of the sequencer. Also adds some helpers in order to access evm storage and nonce.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/ef-testing/src/evm_sequencer/constants.rs
+++ b/crates/ef-testing/src/evm_sequencer/constants.rs
@@ -72,12 +72,14 @@ lazy_static! {
     pub static ref CONTRACT_ACCOUNT_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/contract_account.json").unwrap()).unwrap();
     pub static ref EOA_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/externally_owned_account.json").unwrap()).unwrap();
     pub static ref PROXY_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/proxy.json").unwrap()).unwrap();
+    pub static ref FEE_TOKEN_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/fixtures/ERC20.json").unwrap()).unwrap();
 
     // Main class hashes
     pub static ref KAKAROT_CLASS_HASH: ClassHash = ClassHash(KAKAROT_CLASS.class_hash().unwrap().into());
     pub static ref CONTRACT_ACCOUNT_CLASS_HASH: ClassHash = ClassHash(CONTRACT_ACCOUNT_CLASS.class_hash().unwrap().into());
     pub static ref EOA_CLASS_HASH: ClassHash = ClassHash(EOA_CLASS.class_hash().unwrap().into());
     pub static ref PROXY_CLASS_HASH: ClassHash = ClassHash(PROXY_CLASS.class_hash().unwrap().into());
+    pub static ref FEE_TOKEN_CLASS_HASH: ClassHash = ClassHash(FEE_TOKEN_CLASS.class_hash().unwrap().into());
 
 }
 
--- a/crates/ef-testing/src/evm_sequencer/constants.rs
+++ b/crates/ef-testing/src/evm_sequencer/constants.rs
@@ -9,6 +9,12 @@ use starknet_api::{
     hash::StarkFelt,
 };
 
+fn load_legacy_contract_class(path: &str) -> Result<LegacyContractClass, eyre::Error> {
+    let file = std::fs::File::open(path)?;
+    let class = serde_json::from_reader::<_, LegacyContractClass>(file)?;
+    Ok(class)
+}
+
 lazy_static! {
     // Chain params
     pub static ref CHAIN_ID: u64 = 0x4b4b5254;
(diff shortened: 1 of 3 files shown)


crates/ef-testing/src/evm_sequencer/constants.rs
```diff
@@ -72,12 +72,14 @@ lazy_static! {
     pub static ref CONTRACT_ACCOUNT_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/contract_account.json").unwrap()).unwrap();
     pub static ref EOA_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/externally_owned_account.json").unwrap()).unwrap();
     pub static ref PROXY_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/proxy.json").unwrap()).unwrap();
+    pub static ref FEE_TOKEN_CLASS: LegacyContractClass = serde_json::from_reader::<_, LegacyContractClass>(std::fs::File::open("../../lib/kakarot/build/fixtures/ERC20.json").unwrap()).unwrap();
```
Eikix, line 75: I think we can even do `include_str!` macro, to inline the string at compile time
and use serde_json::from_str 


that is in case the .json isn't too big, we might win a bit of speed;D

Eikix: Otherwise, might be useful to abstract the logic of these LegacyContractClass? such as 
```rust
fn load_legacy_contract_class(path: &str) -> Result<LegacyContractClass, Box<dyn std::error::Error>> {
    let file = std::fs::File::open(path)?;
    let class = serde_json::from_reader::<_, LegacyContractClass>(file)?;
    Ok(class)
}

pub static ref EOA_CLASS: LegacyContractClass = load_legacy_contract_class("../../lib/kakarot/build/externally_owned_account.json").expect("Failed to load EOA_CLASS");
pub static ref PROXY_CLASS: LegacyContractClass = load_legacy_contract_class("../../lib/kakarot/build/proxy.json").expect("Failed to load PROXY_CLASS");
pub static ref FEE_TOKEN_CLASS: LegacyContractClass = load_legacy_contract_class("../../lib/kakarot/build/fixtures/ERC20.json").expect("Failed to load FEE_TOKEN_CLASS");

```

greged93: for include_str!, you mean for all the contract classes? That might be a little big no?

Eikix: maybe a bit too big yeah, you're right

Eikix: but i don't know how/where to find the optimum between binary size vs. runtime overhead of I/O

Eikix (commented):
Nice PR! It's hard to catch up the complexity of the sequencer, I think the whole codebase of the custom sequencer and evm state trait deserves extensive documentation for our culture and maintaibility, this can be a standalone issue to add rustdoc to your newly created sequencer as well as document the steps of internal functions

crates/ef-testing/src/evm_sequencer/constants.rs
greged93: I'm not sure it changes much in this case, if you add `include_str`, binary will contain all this data and load it quickly so `initialize` of the sequencer might win some time (on the first init) but pretty sure that's it

greged93:
Made and assigned issue #469

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -142,12 +143,48 @@ impl EvmState for KakarotSequencer {
...
         for (k, v) in storage {
-            (&mut self.0.state).set_storage_at(starknet_address, k, v);
+            (&mut self.0.state).set_storage_at(*FEE_TOKEN_ADDRESS, k, v);
         }
         Ok(())
     }
+
+    fn get_storage_at(&mut self, evm_address: &Address, key: U256) -> StateResult<U256> {
+        let keys = split_u256(key).map(Into::into);
+        let keys = get_uint256_storage_var_addresses("storage_", &keys).unwrap(); // safe unwrap: all vars are ASCII
```
Eikix, line 154: this cannot work on EOAs right? 
what will happen in case the evm_address from the params is an EOA, or a non-deployed EVM address?

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -142,12 +143,48 @@ impl EvmState for KakarotSequencer {
...
+
+        let nonce_protocol = (&mut self.0.state)
+            .get_nonce_at(starknet_address.try_into()?)?
+            .0;
+
+        let key = get_storage_var_address("nonce", &[])?;
+        let nonce_managed =
+            (&mut self.0.state).get_storage_at(starknet_address.try_into()?, key)?;
+
+        let nonce = if nonce_managed > nonce_protocol {
```
Eikix, line 178: is this safe?
are we sure `nonce_managed > nonce_protocol` <=> account is a CA
and `nonce_managed < nonce_protocol` <=> account is EOA 

?

crates/ef-testing/src/evm_sequencer/mod.rs
```diff
@@ -76,11 +79,25 @@ impl KakarotSequencer {
                     .map_err(|err| StateError::ProgramError(ProgramError::Parse(err)))?,
             )?),
         )?;
+        (&mut self.0.state).set_contract_class(
+            &FEE_TOKEN_CLASS_HASH,
+            ContractClass::V0(ContractClassV0::try_from_json_string(
```
Eikix, line 84: why do we need this?

can't we store FEE_TOKEN_CLASS as a ContractClass already?

crates/ef-testing/src/evm_sequencer/evm_state.rs
greged93: So either both nonces are 0, meaning it will just return 0, either one of the nonce was incremented, meaning this is the used nonce. I can return an error if both nonce are different from 0, would that work?

crates/ef-testing/src/evm_sequencer/mod.rs
greged93: so I'm using this serde conversion because conversion between starknet-rs types and blockifier types is complicated. I need to use the starknet-rs type in order to have access to the `class_hash()` function.

crates/ef-testing/src/evm_sequencer/evm_state.rs
Eikix: I think then we need to document this "proxy" method a bit, to make sure it's readable

-> when a CA is created with CREATE opcode, its protocol nonce is 0 and its managed nonce becomes 1-> this technique works
-> when an EOA is created its nonce is 0, and its managed nonce will be 0 -> this technique works 
-> when an EOA interacts with the chain, its nonce is incremented, and its managed nonce remains 0 -> this technique works
-> when a CA creates more accts, its managed nonce is incremented -> this technique works

crates/ef-testing/src/evm_sequencer/mod.rs
Eikix: I see, very cursed

Eikix (approved):
lgtm

Issue #97, opened 2023-09-11
epic: remove the kakarot-rpc dependency


greged93:
This epic issue will track the different issues that need to be fulfilled in order to remove the `kakarot-rpc` dependancy from ef-tests.

```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/ef-tests/issues/98
- [ ] https://github.com/kkrt-labs/ef-tests/issues/99
```

Eikix:
What is the status of this? Do we need to slightly rename/rework this epic?

greged93:
Might, let's sync with @ClementWalter, I think his main issue was to have a dependency on RPC, but when we refactor the RPC it will just be an SDK. Could we keep open until that refactor?

Issue #99, opened 2023-09-11
Remove kakarot-rpc-code


greged93:
The project currently depends on kakarot-rpc-core for the below functions:
- `split_u256_into_field_elements`: make this a util in ef-tests.

The below traits:
- `KakarotEthApi`, `KakarotStarknetApi`: used to call `KakarotClient.send_transaction` and `KakarotClient.starknet_provider()`. The second can be fixed by having a provider created in the test. The first is more complicated and might require us to perform evm -> starknet transaction conversion in ef-tests.

The below types:
- `Felt252Wrapper`: make this a ef-tests type. This would mean ef-tests becomes the central repo for any commun evm <-> starknet conversions.
- `DeployedKakarot`, `KakarotTestEnvironmentContext`: should be removed once the Katana dumps are generated in `Kakarot` repo.
- `EthApiError`: due to the call to `KakarotClient.send_transaction`. Can be removed if `KakarotEthApi` trait is removed.
- `ConversionError`: due to the import of `Felt252Wrapper`. Can be removed if `Felt252Wrapper` is moved to ef-tests.

And the below constants:
- `CHAIN_ID`: make this a constant in ef-tests.

Issue #98, opened 2023-09-11
Remove hive-utils dependency


greged93:
The project currently depends on hive-utils for the below functions:
- `compute_starknet_address`: move to a util folder in ef-tests.
- `genesis_set_bytecode`, `genesis_set_storage_kakarot_contract_account`, `genesis_approve_kakarot`, `genesis_fund_starknet_address`: replace in ef-tests with types wrapper (around Bytes, Account, Balance,...) and define a StoreStarknet trait which consumes the type in order to return a `Vec<(K, V)>`, where K (key) and V (value) are generics to the trait.

And the below types: 
- `ContractAddress`, `StorageKey` and `StorageValue`: make a type alias in ef-tests.

jobez:
great issue!

Eikix:
concerning first point, I suggest we plan a great refactor of the kakarot-rpc repository.

1. change the scope of the kakarot-rpc repository. it's a de-facto sdk for kakarot. rename the kakarot-rpc to kakarot-rust. it will eventually contain a `kakarot-sdk` crate, which contains all our runtime utils, a `test-utils` crate for test-utils, and a `eth-rpc` crate for the eth-rpc content.
thus, i think compute_starknet_address should live in the `kakarot-sdk` crate, all our test-utils in test-utils, as @bajpai244 is refactoring.

2. good idea for the trait! i completely agree, where do you think it should live? in the ef-tests repo? what if it's re-used in hive? should it be rather under test-utils?

Eikix:
What is the status of this? since we now depending on test-utils, is this fine?

greged93:
I guess we could leave it, but also we had an idea for the test sequencer for ef-tests, where we would wrap it into a EvmSequencer, and have the helpers implemented on this, so will be cleaner. I think we can keep this open and recheck the status after sequencer is done, so maybe end of the week?

Eikix:
What is the status of this?

greged93:
can be closed!

PR #472, opened 2023-10-05
feat: deprecate katana

greged93:
# Pull Request type
Removes the dep on Katana and Kakarot RPC, replacing it with our own custom sequencer.

Time spent on this PR: 2 days

Resolves: #97 #98 #99

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, renaming)
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?
Ef-tests are now ran on our own sequencer.

- [x] Yes
- [ ] No


blockchain-tests-skip.yml
```diff
@@ -304,6 +304,7 @@ filename:
     - CREATE_EContractCreateNEContractInInit_Tr.json # ef-tests #340
     - CREATE_empty000CreateinInitCode_Transaction.json # ef-tests #340
     - CREATE_FirstByte_loop.json # ef-tests #340
+    - CreateTransactionHighNonce.json # ef-tests #471
```
Eikix, line 307: why are we adding this one again?

greged93: This is the one that causes blockifier to crash, due to increasing u64::max nonce by one :/

crates/ef-testing/src/evm_sequencer/constants.rs
```diff
@@ -41,9 +41,9 @@ lazy_static! {
             block_number: BlockNumber(0),
             block_timestamp: BlockTimestamp(0),
             sequencer_address: *SEQUENCER_ADDRESS,
-            fee_token_address: *FEE_TOKEN_ADDRESS,
+            fee_token_addresses: FeeTokenAddresses{eth_fee_token_address: *ETH_FEE_TOKEN_ADDRESS, strk_fee_token_address: *STRK_FEE_TOKEN_ADDRESS},
```
Eikix, line 44: can you add comments explaining these two?

crates/ef-testing/src/evm_sequencer/constants.rs
```diff
@@ -41,9 +41,9 @@ lazy_static! {
...
             vm_resource_fee_cost: Arc::new(VM_RESOURCES.clone()),
-            gas_price: 1,
+            gas_prices: GasPrices{eth_l1_gas_price: 1, strk_l1_gas_price: 1},
```
Eikix, line 46: can you add comment explaining what this does?

crates/ef-testing/src/evm_sequencer/constants.rs
```diff
@@ -59,7 +59,7 @@ lazy_static! {
         ))
         .unwrap()
     );
-    pub static ref FEE_TOKEN_ADDRESS: ContractAddress = ContractAddress(
+    pub static ref ETH_FEE_TOKEN_ADDRESS: ContractAddress = ContractAddress(
```
Eikix, line 62: Never asked before but is there no way to construct the ContractAddress type in a simpler manner than going through 3 conversions?

crates/ef-testing/src/evm_sequencer/constants.rs
```diff
@@ -68,6 +68,15 @@ lazy_static! {
         ))
         .unwrap()
     );
+    pub static ref STRK_FEE_TOKEN_ADDRESS: ContractAddress = ContractAddress(
```
Eikix, line 71: same question

crates/ef-testing/src/evm_sequencer/mod.rs
```diff
@@ -86,15 +86,16 @@ impl KakarotSequencer {
...
+        (&mut self.0.state).set_class_hash_at(*ETH_FEE_TOKEN_ADDRESS, *FEE_TOKEN_CLASS_HASH)?;
 
         Ok(self)
     }
 }
 
 impl Execution for KakarotSequencer {
     fn execute(&mut self, transaction: Transaction) -> Result<(), TransactionExecutionError> {
-        self.0.execute(transaction)
+        let _ = self.0.execute(transaction);
```
Eikix, line 97: why do we need this?

crates/ef-testing/src/evm_sequencer/constants.rs
greged93: This is actually a change made by blockifier to their type, I can but I don't have much context, other then they probably plan for users to be able to pay fees in stark and eth.

crates/ef-testing/src/evm_sequencer/constants.rs
greged93: Same here, blockifier update

crates/ef-testing/src/evm_sequencer/constants.rs
greged93: There is actually a macro for this, will update

crates/ef-testing/src/evm_sequencer/mod.rs
greged93: I used to return the error, which caused the runner to crash on any transaction execution or validation issue. But the way the ef-tests are designed is for any failing transaction to be caught via the post state. So if your tx is supposed to fail, this will be reflected in the post state. An example: the tx I send to the contract will attempt to write 1 in the storage. However, this tx should fail because of XYZ. The post state should then NOT contain 1 in the storage.

Eikix (approved):

Issue #470, opened 2023-10-05
feat: implement balance and bytecode verification


greged93:
## Feature Request
We currently do not assert in the poststate of our runner on the bytecode or the balance of accounts. The balance of the sender can be a little tricky because of the difference in gas but needs to be implemented for all other accounts. The bytecode is required for new created accounts.

### Describe Preferred Solution
Implement getters for the bytecode and the account balance in the evm sequencer. Call these functions in the post state handling of the runner.

Issue #455, opened 2023-09-30
bug: missing field `gasPrice` panic error


khaeljy:
# Bug Report

**Kakarot version:**

Kakarot sha 7686fbf022d99ce01927eb488ad59f0a0ec7ce5d

**Current behavior:**

When running stEIP1559, stEIP2930, stExample

```
thread 'blockchain_tests::st_eip1559' panicked at 'test case should load: Io { path: "/workspaces/ef-tests/crates/ef-testing/ethereum-tests/GeneralStateTests/stEIP1559/intrinsic.json", error: "missing field `gasPrice` at line 1 column 2890321" }', /workspaces/ef-tests/crates/ef-testing/src/traits/mod.rs:56:58

thread 'blockchain_tests::st_eip2930' panicked at 'test case should load: Io { path: "/workspaces/ef-tests/crates/ef-testing/ethereum-tests/GeneralStateTests/stEIP2930/coinbaseT2.json", error: "missing field `gasPrice` at line 1 column 5326" }', /workspaces/ef-tests/crates/ef-testing/src/traits/mod.rs:56:58

thread 'blockchain_tests::st_example' panicked at 'test case should load: Io { path: "/workspaces/ef-tests/crates/ef-testing/ethereum-tests/GeneralStateTests/stExample/basefeeExample.json", error: "missing field `gasPrice` at line 1 column 9702" }', /workspaces/ef-tests/crates/ef-testing/src/traits/mod.rs:56:58
```

greged93:
Will be added in #473, comes from the fact that the structure we deserialize in should have Option as a type for the gas_price field.

greged93:
Completed by https://github.com/kkrt-labs/ef-tests/pull/473

Issue #257, opened 2023-09-22
bug: callWithHighValueAndGasOOG.json error when loading test case file


khaeljy:
# Bug Report

**Kakarot version:**

Kakarot sha 7686fbf022d99ce01927eb488ad59f0a0ec7ce5d

**Current behavior:**

When running stCallCreateCallCodeTest

```
thread 'blockchain_tests::st_call_create_call_code_test' panicked at 'test case should load: Io { path: "/workspaces/ef-tests/crates/ef-testing/ethereum-tests/GeneralStateTests/stCallCreateCallCodeTest/callWithHighValueAndGasOOG.json", error: "invalid length 18, expected a (both 0x-prefixed or not) hex string or byte array containing between (0; 8] bytes at line 1 column 6249" }', /workspaces/ef-tests/crates/ef-testing/src/traits/mod.rs:56:58
stack backtrace:
   0:     0x55ae385f1551 - std::backtrace_rs::backtrace::libunwind::trace::hd102c46c16771587
                               at /rustc/3b2073f0762cff4d3d625bb10017e0ce4e7abe50/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x55ae385f1551 - std::backtrace_rs::backtrace::trace_unsynchronized::h6545cf7d6ed27551
                               at /rustc/3b2073f0762cff4d3d625bb10017e0ce4e7abe50/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55ae385f1551 - std::sys_common::backtrace::_print_fmt::h258fac35947dded0
                               at /rustc/3b2073f0762cff4d3d625bb10017e0ce4e7abe50/library/std/src/sys_common/backtrace.rs:65:5
   3:     0x55ae385f1551 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hf070a48cb28a24d1
                               at /rustc/3b2073f0762cff4d3d625bb10017e0ce4e7abe50/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x55ae3861e14f - core::fmt::rt::Argument::fmt::h0a093c27aabe85c6
                               at /rustc/3b2073f0762cff4d3d625bb10017e0ce4e7abe50/library/core/src/fmt/rt.rs:138:9
   5:     0x55ae3861e14f - core::fmt::write::h809e1a6b668324f6
[...]

PR #473, opened 2023-10-05
test: update post-state assertion

greged93:
# Pull Request type
Updates the post state assertion of the runner in order to add bytecode and balance assertion.

Time spent on this PR: 0.5 day

Resolves: #470 #476 #257 #455

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?
Post state of the runner will now assert on the bytecode and balance of accounts, only skipping the balance for the sender due to the difference in gas metering.

- [ ] Yes
- [x] No


Eikix:
I wonder if these errors in the CI "Error while executing a hint" count as new discovered problems in our EVM or unrelated

crates/ef-testing/src/utils/mod.rs
```diff
@@ -20,3 +23,12 @@ pub(crate) fn deserialize_into<T: for<'a> Deserialize<'a>>(
         error: error.to_string(),
     })
 }
+
+pub(crate) fn address_from_private_key(sk: B256) -> Result<Address, RunnerError> {
```
Eikix, line 27: does ether-rs have a similar method?

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -191,6 +197,50 @@ impl EvmState for KakarotSequencer {
...
+        bytecode.append(&mut felt_to_bytes(&code.into(), remainder as usize).to_vec());
+
+        Ok(Bytes::from(bytecode))
+    }
+
+    fn get_balance_at(&mut self, evm_address: &Address) -> StateResult<U256> {
+        let starknet_address = compute_starknet_address(evm_address);
+
+        let balance_keys = get_erc20_balance_var_addresses(&starknet_address.try_into()?)?;
+        let balance_keys = [balance_keys.0, balance_keys.1];
```
Eikix, line 234: no need to put it in an array

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -191,6 +197,50 @@ impl EvmState for KakarotSequencer {
...
+
+        let remainder = bytecode_len % 16;
+        let key = get_storage_var_address("bytecode_", &[StarkFelt::from(full_16)])?;
+        let code = (&mut self.0.state).get_storage_at(starknet_address.try_into()?, key)?;
+        bytecode.append(&mut felt_to_bytes(&code.into(), remainder as usize).to_vec());
+
+        Ok(Bytes::from(bytecode))
+    }
+
+    fn get_balance_at(&mut self, evm_address: &Address) -> StateResult<U256> {
```
Eikix, line 230: maybe add a rustdoc to say this gets the balance of the native token (starknet_eth) of the underlying chain

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -191,6 +197,50 @@ impl EvmState for KakarotSequencer {
             Into::<FieldElement>::into(nonce).to_bytes_be(),
         ))
     }
+
+    fn get_code_at(&mut self, evm_address: &Address) -> StateResult<Bytes> {
+        let starknet_address = compute_starknet_address(evm_address);
+
+        let bytecode_len = (&mut self.0.state).get_storage_at(
```
Eikix, line 204: will this be 0 for an EOA?

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -191,6 +197,50 @@ impl EvmState for KakarotSequencer {
             Into::<FieldElement>::into(nonce).to_bytes_be(),
         ))
     }
+
+    fn get_code_at(&mut self, evm_address: &Address) -> StateResult<Bytes> {
```
Eikix, line 201: how can we know this works?

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -191,6 +197,50 @@ impl EvmState for KakarotSequencer {
...
+        let bytecode_len = (&mut self.0.state).get_storage_at(
+            starknet_address.try_into()?,
+            get_storage_var_address("bytecode_len_", &[])?,
+        )?;
+        let bytecode_len: u64 = bytecode_len.try_into()?;
+        if bytecode_len == 0 {
+            return Ok(Bytes::default());
+        }
+
+        let full_16 = bytecode_len / 16;
```
Eikix, line 213: slightly confusing name, can you find something else?


can we also add comments?

-> we're assuming here that we pack 16 bytes of bytecode in the CA (which is ok, just need to document it imo, it'll change for ssj)

crates/ef-testing/src/evm_sequencer/evm_state.rs
greged93: Yes, the EOA will not have the storage var set for bytecode_len_, hence this will always return 0 for an eoa

crates/ef-testing/src/evm_sequencer/evm_state.rs
greged93: Running `cargo test` runs the add.json and mul.json files from the ef-tests folder in the VMTests suite. If this wouldn't work, we would have issues there.

crates/ef-testing/src/evm_sequencer/evm_state.rs
```diff
@@ -210,23 +210,25 @@ impl EvmState for KakarotSequencer {
             return Ok(Bytes::default());
         }
 
-        let full_16 = bytecode_len / 16;
+        // Assumes that the bytecode is stored in 16 byte chunks.
+        let len_16_bytes_chunk = bytecode_len / 16;
```
Eikix, line 214:
```suggestion
        let num_chunks = bytecode_len / 16;
```

crates/ef-testing/src/evm_sequencer/utils.rs
```diff
@@ -47,6 +47,10 @@ pub fn bytes_to_felt_vec(bytes: &Bytes) -> Vec<FieldElement> {
     bytes.to_vec().into_iter().map(FieldElement::from).collect()
 }
 
+pub fn felt_to_bytes(felt: &FieldElement, len: usize) -> Bytes {
+    Bytes::from(&felt.to_bytes_be()[16..len + 16])
```
ClementWalter, line 51: why do we slice from `16`?

crates/ef-testing/src/models/case.rs
```diff
@@ -175,13 +179,34 @@ impl BlockchainTestCase {
...
             }
+            // Bytecode
+            let actual = sequencer.get_code_at(address)?;
+            if actual != expected_state.code {
+                return Err(RunnerError::Other(format!(
+                    "code mismatch for {:#20x}: expected {:#x}, got {:#x}",
+                    address, expected_state.code, actual
+                )));
+            }
+            // Skip sender address because of the difference in gas cost
```
ClementWalter, line 198: can't we force gas price = 0 instead?

crates/ef-testing/src/evm_sequencer/utils.rs
greged93: This comes from the way we implemented bytecode in Kakarot, we only store on 16 bytes. But I see what you mean, I can rename to `high_16_bytes_felt_to_bytes`.

crates/ef-testing/src/models/case.rs
greged93: I can force gas price to 0 in the Kakarot sequencer but the ef test will still include gas cost in the evm, meaning I will always have a difference.

ClementWalter: indeed, and then we update the test definition for the gas_price only

ClementWalter (approved):

PR #525, opened 2023-10-11
Unskip passing and add newly failing due to GAS

ClementWalter:
The 2 added tests were probably passing before due to a VM error and consequently no state change.
After the return_data fix, the Cairo VM doesn't revert anymore and the tests are failing because of no proper gas accounting.


Eikix (approved):
lgtm

PR #539, opened 2023-10-12
Skip REVERT involding balance

ClementWalter:
See https://github.com/kkrt-labs/kakarot/pull/769

The REVERT refacto is still partial and needs to handle funds. To make the PR not too big, let's skip these tests first, and unskip once it's fixed.


Eikix (approved):
lgtm

PR #547, opened 2023-10-17
test: load macro

greged93:
# Pull Request type

Time spent on this PR: 3 days

Resolves: #545 

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [x] Testing

## What is the new behavior?
Instead of running one test per ef-tests folder (hence about 60 tests), we now run one tests per test subcase of each file (hence +19k tests). Tests that are skipped are also marked as `ignored`, which allows for better tracking of the skipped tests.

- [x] Yes
- [ ] No


Cargo.toml
```diff
@@ -1,5 +1,9 @@
 [workspace]
-members = ["crates/ef-testing", "crates/sequencer"]
+members = [
+  "crates/ef-testing",
```
Eikix, line 3:
```suggestion
  "crates/*",
```
does this work in rust?

greged93: I think it does, didn't know, nice!

Makefile
```diff
@@ -41,8 +41,8 @@ unit:
 
 # Runs the repo tests
 tests:
-	cargo nextest run --failure-output final
+	cargo nextest run --failure-output final --no-fail-fast --final-status-level fail
```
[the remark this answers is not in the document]
greged93: this will only log the output at the end of the tests, for failing test only and won't stop on the first failing test, running ALL tests instead.

crates/ef-testing-proc-macro/src/constants.rs
```diff
@@ -0,0 +1,9 @@
+use std::path::{Path, PathBuf};
+
+lazy_static::lazy_static! {
+    pub(crate) static ref TEST_FILTER_PATH: PathBuf =  Path::new("blockchain-tests-skip.yml").to_path_buf();
+    pub(crate) static ref REGEX: regex::Regex = regex::Regex::new(r"[=^.]").unwrap();
```
Eikix, line 5: why do we need this? 
the name isn't so descriptive, maybe we should rename it

crates/ef-testing-proc-macro/src/utils.rs
```diff
@@ -0,0 +1,32 @@
+use std::path::{Path, PathBuf};
+
+use crate::dir_reader::PathWrapper;
+
+/// Converts the path to a string and removes the "BlockchainTests" folder
```
Eikix, line 5: this isn't so clear, can you provide an example in the rustdoc?

crates/ef-testing-proc-macro/src/utils.rs
```diff
@@ -0,0 +1,32 @@
...
+pub(crate) fn blockchain_tests_to_general_state_tests_path(path: PathWrapper) -> PathWrapper {
+    Into::<PathBuf>::into(path)
+        .components()
+        .filter(|x| !x.as_os_str().eq_ignore_ascii_case("BlockchainTests"))
+        .collect::<PathBuf>()
+        .into()
+}
+
+/// Converts the path to a Vector of Strings
+pub(crate) fn path_to_vec_string(path: &Path) -> Result<Vec<String>, eyre::Error> {
```
Eikix, line 15: why do we need to go from having Paths to having vectors of strings?

crates/ef-testing-proc-macro/src/utils.rs
```diff
@@ -0,0 +1,32 @@
...
+                .ok_or_else(|| eyre::eyre!("Error while converting the path to a string"))?
+                .to_string())
+        })
+        .collect()
+}
+
+/// Trims the path at the given folder, returning only the path after the folder
+pub(crate) fn trim_path_at_folder(path: Vec<String>, folder: &str) -> Vec<String> {
+    path.into_iter()
+        .skip_while(|x| x != folder)
```
Eikix, line 29: we're able to compare a String and a &str?

crates/ef-testing-proc-macro/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "ef-testing-proc-macro"
```
Eikix, line 2: this name is not really expressive imho,
i don't think we should name a crate after what its main goal/ its main attribute

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
+use std::collections::BTreeMap;
+
+use serde_json::Value;
+
+use crate::{dir_reader::PathWrapper, utils::blockchain_tests_to_general_state_tests_path};
+
+pub struct ContentReader;
+
+impl ContentReader {
+    pub fn secret_key(path: PathWrapper) -> Result<Value, eyre::Error> {
```
Eikix, line 10: I think this deserves to be documented,

as I understand it, our tests' paths are `BlockchainTests/GeneralStateTests/**/someTest.json` and the secret key is located at `GeneralStateTests/**/someTest.json`

we should provide example and explanation

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
...
+        let cases: BTreeMap<String, Value> = serde_json::from_str(&content)?;
+        let case = cases
+            .into_values()
+            .collect::<Vec<_>>()
+            .first()
+            .ok_or_else(|| eyre::eyre!("No case found"))?
+            .clone();
+
+        Ok(case
+            .get("transaction")
```
Eikix, line 23: can it happen that there is more than one transaction key?

crates/ef-testing-proc-macro/src/constants.rs
greged93: true, the regexp is used to filter out characters that appear in the ef-tests but are not valid for rust identifiers.

crates/ef-testing-proc-macro/src/utils.rs
greged93: The issue I ran into was that it was a lot easier to handle a Vec<String> over a Path: using a Vec<String>, I can pop the first value in the array, store that in my `DirReader` structure and keep recursing. However, a `Path` doesn't have a pop_front method, and I was stuck with some complicated helpers. I found the method of just using `Vec<String>` easier.

crates/ef-testing-proc-macro/src/utils.rs
greged93: pretty sure the compiler auto-dereferences

crates/ef-testing-proc-macro/Cargo.toml
greged93: I followed the naming from cairo corelib for proc-macros: https://github.com/starkware-libs/cairo/tree/main/crates/cairo-lang-proc-macros

crates/ef-testing-proc-macro/src/content_reader.rs
greged93: I don't think you have a valid json if you have more than once the same key right?

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub struct ContentReader;
+
+impl ContentReader {
+    pub fn secret_key(path: PathWrapper) -> Result<Value, eyre::Error> {
+        let path = blockchain_tests_to_general_state_tests_path(path);
+        let content = path.read_file_to_string()?;
+
+        let cases: BTreeMap<String, Value> = serde_json::from_str(&content)?;
+        let case = cases
```
Eikix, line 15: Mhm interestingly enough, tests cases in GeneralStateTests seem to only ever have one case^^, and have nested keys for forks, mhm

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
...
+            .into_values()
+            .collect::<Vec<_>>()
+            .first()
+            .ok_or_else(|| eyre::eyre!("No case found"))?
+            .clone();
```
Eikix, line 20: how about
`let case = cases.into_values().next().ok_or_else(|| eyre::eyre!("No case found"))?;`

crates/ef-testing-proc-macro/src/utils.rs
Eikix: I'm surprised there is no known library to recursively explore a directory and create a sort of tree for all files inside it

crates/ef-testing-proc-macro/Cargo.toml
Eikix: mhmm ok but here we'll only ever have one, no?

crates/ef-testing-proc-macro/src/content_reader.rs
Eikix: I guess then the question becomes, is there a situation where there are multiple cases in any given file in GeneralStateTests?

crates/ef-testing-proc-macro/src/content_reader.rs
greged93: I think that works, nice

crates/ef-testing-proc-macro/src/content_reader.rs
greged93: There might be, but since we are only taking the transaction to extract the secret key, there hasn't been an issue up to now.

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
...
+        Ok(test_case
+            .get("postState")
+            .ok_or_else(|| eyre::eyre!("No post state found"))?
+            .clone())
+    }
+
+    pub fn block(test_case: &Value) -> Result<Value, eyre::Error> {
+        Ok(test_case
+            .get("blocks")
+            .and_then(|blocks| blocks.as_array().map(|blocks| blocks.first()))
```
Eikix, line 49: are we sure there's only ever one block? 
will this change? we might have to add a comment to follow up

Eikix: PS: 
What should we prefer here between conciseness and Rust mastery or readability?

```rust
pub fn block(test_case: &Value) -> Result<Value, eyre::Error> {
    // Attempt to get the "blocks" value
    let blocks = test_case
        .get("blocks")
        .ok_or_else(|| eyre::eyre!("Key 'blocks' not found"))?;
    
    // Ensure it's an array
    let blocks_array = blocks
        .as_array()
        .ok_or_else(|| eyre::eyre!("'blocks' is not an array"))?;
    
    // Get the first block
    let first_block = blocks_array
        .first()
        .ok_or_else(|| eyre::eyre!("'blocks' array is empty"))?;

    // Return a clone of the block
    Ok(first_block.clone())
}
```

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
+use std::collections::BTreeMap;
+
+use serde_json::Value;
+
+use crate::{dir_reader::PathWrapper, utils::blockchain_tests_to_general_state_tests_path};
+
+pub struct ContentReader;
+
+impl ContentReader {
```
Eikix, line 9: Please add rustdoc to express that this content reader is meant for one particular test case, inside a LotsOfTestCases.json file,
give example of structure

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
...
+
+        Ok(case
+            .get("transaction")
+            .ok_or_else(|| eyre::eyre!("No transaction found"))?
+            .get("secretKey")
+            .ok_or_else(|| eyre::eyre!("No secret key found"))?
+            .clone())
+    }
+
+    pub fn pre_state(test_case: &Value) -> Result<Value, eyre::Error> {
```
Eikix, line 30: Nice!

crates/ef-testing-proc-macro/src/utils.rs
greged93: You can walk a directory but I was not able to find a known crate that allows you to recursively construct a mapping of the folders.

crates/ef-testing-proc-macro/src/content_reader.rs
```diff
@@ -0,0 +1,54 @@
...
+        Ok(serde_json::from_value(
+            test_case
+                .get("pre")
+                .ok_or_else(|| eyre::eyre!("No pre state found"))?
+                .clone(),
+        )?)
+    }
+
+    pub fn post_state(test_case: &Value) -> Result<Value, eyre::Error> {
```
Eikix, line 39: Thinking out loud, I'm okay with eyre for this PR, but it's preferable to use custom errors with thiserror and use eyre only in the highest function, such as main.rs, it's quite compact to write an error enum imo

greged93: I would also normally use thiserror, but since all the functions here are only used to generate the code for the macro, it felt wasteful to have a real nice error enum when there is little error handling going on.

crates/ef-testing-proc-macro/src/content_reader.rs
greged93: I'm not sure there will always be one block, I have not seen a situation with more than one block, we can update if that happens when debugging the tests.

I tried just to be concise for this function as it doesn't really perform anything interesting, just reads the value from a json, but can update to this if you want.

crates/ef-testing-proc-macro/src/converter.rs
```diff
@@ -0,0 +1,139 @@
...
+use serde_json::Value;
+
+use crate::{
+    constants::{FORK, REGEX},
+    content_reader::ContentReader,
+    dir_reader::{DirReader, PathWrapper},
+};
+
+/// The TestConverter is used to convert all the
+/// tests files in to rust tests.
```
Eikix, line 13: I don't understand what this means, can this be more specific?

greged93: I plan to update to this, is this ok?
```
/// The TestConverter is used to convert the directory structure
/// into a String containing all the rust tests to be ran.
///
/// # Example
///
/// Test location: BlockchainTests/GeneralStateTests/stRandom/
/// List of tests: [randomStatetest0.json, randomStatetest1.json, ...]
/// Inner tests: [randomStatetest0_d0g0v0_Shanghai, randomStatetest0_d1g0v0_Shanghai,
/// ..., randomStatetest1_d0g0v0_Shanghai, randomStatetest1_d1g0v0_Shanghai, ...]
/// Generated String:
/// r#"
/// mod randomStatetest0 {
///   use super::*;
///   #[test]
///   fn test_randomStatetest0_d0g0v0_Shanghai() {
///     ...
///   }
///   #[test]
///   fn test_randomStatetest0_d1g0v0_Shanghai() {
///     ...
///   }
///   #[test]
///   fn test_randomStatetest1_d0g0v0_Shanghai() {
///     ...
///   }
///   #[test]
///   fn test_randomStatetest1_d1g0v0_Shanghai() {
///     ...
///   }
///   ...
/// }
/// "#
```

Eikix: This is great

crates/ef-testing-proc-macro/src/content_reader.rs
Eikix: Ah I see, how do other people do?

crates/ef-testing-proc-macro/src/content_reader.rs
Eikix: I think it's fine then

crates/ef-testing-proc-macro/Cargo.toml
Eikix: okay lgtm

crates/ef-testing-proc-macro/src/utils.rs
Eikix: Do we need to go from path to vec<string> to accomplish our goal (which is to recursively find all tests?)
could we make it work with:
https://github.com/paradigmxyz/reth/blob/main/testing/ef-tests/src/suite.rs#L50

crates/ef-testing/src/models/case.rs
```diff
@@ -1,109 +1,57 @@
...
-
-        if !should_skip && SKIP.regex.contains_key(dir) {
-            should_skip = SKIP
-                .regex
-                .get(dir)
-                .unwrap()
-                .iter()
-                .any(|regex| Regex::new(regex.as_str()).unwrap().is_match(name));
+    pub fn new(
+        case_name: String,
```
ClementWalter, line 38: why do you change the name from `case_name` to `name`?

crates/ef-testing-proc-macro/src/utils.rs
greged93: I already use Walkdir to find all the file, this doesn't really fix the main issue, which is that I wasn't able to find an easy way to just `pop_front` the first folder in a `Pathbuf`. I did use a solution with Pathbuf at the beginning but using a `Vec<String>` removed 50% of the code, which was mainly doing `.iter().take(1).collect()` everywhere and also `iter().skip(1).collect()`.

crates/ef-testing/src/models/case.rs
```diff
@@ -116,23 +64,11 @@ impl BlockchainTestCase {
...
-        let test = self.test(test_case_name)?;
-
+    fn handle_transaction(&self, sequencer: &mut KakarotSequencer) -> Result<(), RunnerError> {
         // we extract the transaction from the block
-        let block = test
-            .blocks
-            .first()
-            .ok_or_else(|| RunnerError::Other("test has no blocks".to_string()))?;
+        let block = &self.block;
         // we adjust the rlp to correspond with our currently hardcoded CHAIN_ID
```
ClementWalter, line 70: I can update the KKT0 to have the chain id set in constructor and in a storage var

greged93: and then the EOA would make a call to it to get the chain id?

ClementWalter: why not in the initialize, let's do this in another PR

crates/ef-testing/Cargo.toml
```diff
@@ -52,3 +50,4 @@ zip = { workspace = true, optional = true }
 
 [dev-dependencies]
 tracing-subscriber = { workspace = true }
+ef-testing-proc-macro = { path = "../ef-testing-proc-macro" }
```
ClementWalter, line 53: is it a dev deps? a deps no? since it's required not to dev on the repo, but to run it

greged93: it's only required in the integration tests, so setting it as a dev dep works no?

crates/ef-testing-proc-macro/Cargo.toml
ClementWalter: cf also https://doc.rust-lang.org/book/ch19-06-macros.html

> When creating procedural macros, the definitions must reside in their own crate with a special crate type. This is for complex technical reasons that we hope to eliminate in the future.

crates/ef-testing-proc-macro/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version.workspace = true
+edition.workspace = true
+authors.workspace = true
+description.workspace = true
+homepage.workspace = true
+repository.workspace = true
+readme.workspace = true
+license.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
ClementWalter, line 12: can me remove this?

crates/ef-testing-proc-macro/Cargo.toml
ClementWalter, line 1: we sometimes use `package.workspace = true` and sometimes `packages = { workspace = true}`. Is there any way to have clippy enforcing one or another?

greged93: That's my bad, didn't realize the manifest created with `cargo` used package.workspace

crates/ef-testing-proc-macro/src/utils.rs
```diff
@@ -0,0 +1,37 @@
...
+        .components()
+        .filter(|x| !x.as_os_str().eq_ignore_ascii_case("BlockchainTests"))
+        .collect::<PathBuf>()
+        .into()
+}
+
+/// Converts the path to a Vector of Strings
+pub(crate) fn path_to_vec_string(path: &Path) -> Result<Vec<String>, eyre::Error> {
+    path.iter()
+        .map(|os_str| {
```
ClementWalter, line 22: this is more like `relative_to`, for example in python https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.relative_to

crates/ef-testing-proc-macro/src/dir_reader.rs
```diff
@@ -0,0 +1,112 @@
...
+            let root_name = current_path.first().cloned().unwrap(); // safe unwrap
+            let sub_node = self.sub_dirs.entry(root_name).or_insert_with(|| DirReader {
+                sub_dirs: BTreeMap::default(),
+                files: Vec::default(),
+                filter: self.filter.clone(),
+            });
+            sub_node.insert_file(current_path.into_iter().skip(1).collect(), full_path);
+        } else {
+            let skip = self.filter.is_skipped(&full_path);
+            self.files.push((full_path, skip));
```
ClementWalter, line 69: looks like this could be in its own file

crates/ef-testing-proc-macro/src/dir_reader.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct PathWrapper(PathBuf);
+
+impl From<PathBuf> for PathWrapper {
+    fn from(path: PathBuf) -> Self {
+        Self(path)
+    }
+}
```
ClementWalter, line 81: only one is enough no?

crates/ef-testing-proc-macro/src/converter.rs
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /// Converts the given directory into a String.
+    fn convert_folders(node: &DirReader) -> Result<String, eyre::Error> {
+        let mut acc = String::new();
+        for (dir_name, sub_node) in &node.sub_dirs {
+            acc += &TestConverter::format_to_module(dir_name);
+            acc += &Self::convert_folders(sub_node)?;
+            acc += "}";
+        }
```
ClementWalter, line 79: secret_key

ClementWalter (changes requested):
We just need to make sure that a change in `blockchain-tests-skip.yml` will be seen by the compiler, and regenerate the tests accordingly

crates/ef-testing-proc-macro/Cargo.toml
greged93: I am not sure, actually we only use the `.workspace` for `readme`, `license`, etc, which is generated by `cargo init`.

crates/ef-testing-proc-macro/src/dir_reader.rs
greged93: I need `From<PathWrapper> for PathBuf` in utils and dir_reader, so need to keep it. Also need `From<PathBuf> for PathWrapper` for converting when using `WalkDir`.

ClementWalter (approved):

PR #555, opened 2023-10-19
feat: add dumping and loading of state

bajpai244:
# Pull Request type

Time spent on this PR:

Resolves: #159 

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?

Allows dump and loading state to and from json files.

- [ ] Yes
- [x] No


crates/sequencer/Cargo.toml
```diff
@@ -13,13 +13,16 @@ license.workspace = true
 # Starknet
 # TODO: remove the blockifier patch on the workspace once we can remove Katana.
 blockifier = { workspace = true }
+serde.workspace = { workspace = true }
+serde_json.workspace = true
```
greged93, line 17: can we keep one syntax for the workspace?

crates/sequencer/Cargo.toml
```diff
@@ -13,13 +13,16 @@ license.workspace = true
...
 starknet_api = { workspace = true }
 starknet = { workspace = true }
 
 # Other
 eyre = { workspace = true }
 tracing = { workspace = true }
 rustc-hash = "1.1.0"
+thiserror.workspace = true
```
greged93, line 25: same

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,101 @@
...
+    IoError {
+        reason: String,
+        context: std::io::Error,
+    },
+    #[error("{reason:?}")]
+    SerdeJsonError {
+        reason: String,
+        context: serde_json::Error,
+    },
+}
```
greged93, line 26:
```suggestion
pub enum SerializationError {
    #[error(transparent)]
    IoError(#[from] io::Error),
    #[error(transparent)]
    SerdeJsonError(#[from] serde_json::Error),
}
```

crates/sequencer/src/state.rs
```diff
@@ -14,28 +17,58 @@ use starknet_api::{
...
-    classes: FxHashMap<ClassHash, ContractClass>,
-    compiled_class_hashes: FxHashMap<ClassHash, CompiledClassHash>,
-    contracts: FxHashMap<ContractAddress, ClassHash>,
-    storage: FxHashMap<ContractStorageKey, StarkFelt>,
-    nonces: FxHashMap<ContractAddress, Nonce>,
+    pub classes: FxHashMap<ClassHash, ContractClass>,
+    pub compiled_class_hashes: FxHashMap<ClassHash, CompiledClassHash>,
+    pub contracts: FxHashMap<ContractAddress, ClassHash>,
+    pub storage: FxHashMap<ContractStorageKey, StarkFelt>,
+    pub nonces: FxHashMap<ContractAddress, Nonce>,
```
greged93, line 37: mmmm do we need to expose all this? is pub(crate) not enough?

crates/sequencer/src/state.rs
```diff
@@ -14,28 +17,58 @@ use starknet_api::{
...
 }
 
 impl State {
     /// Helper function allowing to set the nonce of a contract.
     pub fn set_nonce(&mut self, contract_address: ContractAddress, nonce: Nonce) {
         self.nonces.insert(contract_address, nonce);
     }
+
+    /// This will serialize the current state, and will save it to a path
+    pub fn dump_state_to_file(&self, path: &PathBuf) -> Result<(), SerializationError> {
```
greged93, line 47: Otherwise, if the reason why you implemented From<&State> because of this, can't you just have `self` instead of `&self`?

crates/sequencer/src/state.rs
```diff
@@ -277,4 +325,56 @@ mod tests {
...
+
+        assert_eq!(state, loaded_state);
+
+        fs::remove_file(&dump_file_path).unwrap_or_else(|error| {
+            panic!(
+                "error in removing file from path {:?},\n error {}",
+                dump_file_path, error
+            )
+        });
+    }
```
greged93, line 379: you can use the [tempfile](https://crates.io/crates/tempfile) crate for this

crates/sequencer/src/state.rs
```diff
@@ -277,4 +325,56 @@ mod tests {
...
+                    dump_file_path, error
+                )
+            });
+
+        let loaded_state = State::load_state_from_file(&dump_file_path).unwrap_or_else(|error| {
+            panic!(
+                "failed loading state from path {:?},\n error {}",
+                dump_file_path, error
+            )
+        });
```
greged93, line 369: I think it would be fine to `expect` here

greged93:
Also, can you integrate all this in the actual execution of the tests?

crates/sequencer/Cargo.toml
bajpai244: fixed via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/b563ec5e17d1ffdc8663505a6880834dcdde40bb) commit.

crates/sequencer/src/serde.rs
bajpai244: resolve via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/b33320f13f052d5f8bfedf2299ef3c30d412a056) commit.

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,101 @@
...
+    fn from(state: &State) -> Self {
+        Self {
+            classes: state.classes.clone(),
+            compiled_classes_hash: state.compiled_class_hashes.clone(),
+            contracts: state.contracts.clone(),
+            storage: state.storage.clone(),
+            nonces: state.nonces.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
bajpai244: Makes sense, I did this to avoid cloning, but now I realize that at the end of the day we seem to be cloning the fields anyway!

Lemme change the implementation

crates/sequencer/src/state.rs
bajpai244: I think we should avoid passsing self as much as possible right?

Because, I think it moves a value { or you clone it }.

Anyway, I have changed the `From` trait implementation.

greged93: I think self is ok if you need to and it removes a clone but in this case I think we can't avoid clone anyway

crates/sequencer/src/state.rs
greged93: right right, I mean more in the sense do we need to format the error with the file path + this extra text? for me just `.expect("failed loading state from path") should give enough info.

crates/sequencer/src/serde.rs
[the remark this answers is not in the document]
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/4d5fb42d2758e4aa15fd339359ea19e4e27a0039) commit.

crates/sequencer/src/state.rs
bajpai244: Gotcha, lemme refactor then!

crates/sequencer/src/state.rs
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/5484da1997d8dbbe67c69ce05d0a7579b6106d94) commit.

crates/sequencer/src/state.rs
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/5484da1997d8dbbe67c69ce05d0a7579b6106d94) commit.

crates/sequencer/src/state.rs
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/47a3e76278392d01c00a97ffed65d9b34d1f84d0) commit.

bajpai244:
> Also, can you integrate all this in the actual execution of the tests?

resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/0ed55db05da8cf5637b2e8774a7bd11c93e935e0).

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,106 @@
...
+    ) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut transformed: HashMap<String, String> = HashMap::new();
+        for (contract_storage_key, storage_value) in map.iter() {
+            let key_str = serde_json::to_string(&contract_storage_key).unwrap();
+            let value_str = serde_json::to_string(&storage_value).unwrap();
+            transformed.insert(key_str, value_str);
+        }
```
greged93, line 74: Mmmmm do we need to make the intermediate mapping? I feel like we could just serialize each (k,v) pair straight into the serializer, skipping this temp value.
Found this which could be of use
https://stackoverflow.com/questions/51276896/how-do-i-use-serde-to-serialize-a-hashmap-with-structs-as-keys-to-json
I think there should be an easier way to do this where we also don't need to implement our own deser method.

greged93: Basically the issue is that you need the key to be a string, but then when you put it into a map and serialize this, what you get is actually a serialize string. Then when serde tries to deserialize the key into a tuple, it fails because the key is now a serialized string. I think we can improve this.

crates/sequencer/src/state.rs
```diff
@@ -277,4 +319,43 @@ mod tests {
...
+        state.classes.insert(class_hash, contract_class);
+        state
+            .compiled_class_hashes
+            .insert(class_hash, compiled_class_hash);
+        state.contracts.insert(contract_address, class_hash);
+        state.storage.insert(contract_storage_key, storage_value);
+        state.nonces.insert(contract_address, nonce);
+
+        let temp_file = tempfile::NamedTempFile::new().expect("failed open named temp file");
+        let dump_file_path = temp_file.into_temp_path();
```
greged93, line 348: nice!

greged93 (commented):
I think there is some improvement that we can bring to the serde module in order to maybe only implement serialization

Eikix:
What is the status of this @bajpai244 @greged93

greged93:
CI failing and waiting for answer on comments

crates/sequencer/Cargo.toml
```diff
@@ -13,14 +13,17 @@ license.workspace = true
...
 # Other
 eyre = { workspace = true }
 tracing = { workspace = true }
 rustc-hash = "1.1.0"
+thiserror = { workspace = true }
 
 [dev-dependencies]
 lazy_static = { workspace = true }
-serde_json = { workspace = true }
+tempfile = "3"
```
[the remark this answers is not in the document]
bajpai244: fixed via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/87faaf2a7fbb66acddf7b3220eb6feb46e486c4c) commit.

crates/sequencer/src/sequencer.rs
```diff
@@ -23,6 +29,7 @@ use tracing::{info, trace, warn};
 pub struct Sequencer<S>
```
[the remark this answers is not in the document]
bajpai244: So, if I understand the reasoning behind this, then it seems to look like this should allow us to have sequencers with states which can or cannot implement the DumpLoad trait.

This way, when a state a would implement Dump and Load, the Sequencer would implement the wrapper methods as well!

Pretty cool! Lemme know if I got the reasoning right!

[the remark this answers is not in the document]
bajpai244: resolved via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/af6efaef4377ebddef97ab29ed6ace89f1a44342) commit.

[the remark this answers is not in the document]
greged93: That's correct :)

crates/sequencer/src/serde.rs
bajpai244: reduced the complexity via [this](https://github.com/kkrt-labs/ef-tests/pull/555/commits/3d27d5b9aa81e1ce48db4c12aba286d6eb362ff7) commit.

greged93:
@bajpai244 you have a couple of issues in the trunk CI you need to fix

Eikix:
I think we shouldn't have any diff on ef test folder

crates/sequencer/src/sequencer.rs
```diff
@@ -43,6 +49,24 @@ where
     }
 }
 
+impl<S> Sequencer<S>
+where
+    S: DumpLoad + Clone,
+    for<'any> &'any mut S: State + StateReader,
+{
+    pub fn dump_state_to_file(&self, file_path: &Path) -> Result<(), SerializationError> {
+        self.state.clone().dump_state_to_file(file_path)
```
Eikix, line 58: why do we need to clone state? isn't it big and lengthy?
maybe this method should consume self

crates/sequencer/src/sequencer.rs
```diff
@@ -43,6 +49,24 @@ where
     }
 }
 
+impl<S> Sequencer<S>
+where
+    S: DumpLoad + Clone,
+    for<'any> &'any mut S: State + StateReader,
```
Eikix, line 55: why do we need this higher rank lifetime?

crates/sequencer/src/state.rs
```diff
@@ -22,13 +24,13 @@ use crate::commit::Committer;
...
 /// See [rustc-hash](https://crates.io/crates/rustc-hash) for more information.
-#[derive(Default)]
+#[derive(Clone, Default, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct State {
-    classes: FxHashMap<ClassHash, ContractClass>,
-    compiled_class_hashes: FxHashMap<ClassHash, CompiledClassHash>,
-    contracts: FxHashMap<ContractAddress, ClassHash>,
-    storage: FxHashMap<ContractStorageKey, StarkFelt>,
-    nonces: FxHashMap<ContractAddress, Nonce>,
+    pub(crate) classes: FxHashMap<ClassHash, ContractClass>,
```
Eikix, line 29: why make this pub crate?

crates/sequencer/src/sequencer.rs
greged93: It's required by the Sequencer

crates/sequencer/src/sequencer.rs
greged93: true !

crates/sequencer/src/state.rs
greged93: pretty sure you need to make it at least pub(crate) if you want to serialize

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,222 @@
...
+        Self: Sized;
+}
+
+impl DumpLoad for State {
+    /// This will serialize the current state, and will save it to a path
+    fn dump_state_to_file(self, path: &Path) -> Result<(), SerializationError> {
+        let serializable_state: SerializableState = self.into();
+
+        let dump = serde_json::to_string(&serializable_state)
+            .map_err(SerializationError::SerdeJsonError)?;
```
Eikix, line 30: Why do we need a map err? it should "just work"

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        fs::write(path, dump).map_err(SerializationError::IoError)?;
```
Eikix, line 32: here as well

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,222 @@
...
+
+        Ok(())
+    }
+
+    /// This will read a dump from a file and initialize the state from it
+    fn load_state_from_file(path: &Path) -> Result<Self, SerializationError> {
+        let dump = fs::read(path).unwrap();
```
Eikix, line 39: shouldn't we let the transparent IO error bubble up?

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,222 @@
...
+        let serializable_state: SerializableState =
+            serde_json::from_slice(&dump).map_err(SerializationError::SerdeJsonError)?;
```
Eikix, line 41: no need for map_err afaik

crates/sequencer/src/serde.rs
```diff
@@ -0,0 +1,222 @@
...
+            contracts: serializable_state.contracts,
+            storage: serializable_state.storage,
+            nonces: serializable_state.nonces,
+        }
+    }
+}
+
+mod serialize_contract_storage {
+    use blockifier::state::cached_state::ContractStorageKey;
+    use rustc_hash::{FxHashMap, FxHasher};
```
Eikix, line 91: why do we need rust C hash data structures?

crates/sequencer/src/state.rs
Eikix: ah i see

crates/sequencer/src/serde.rs
bajpai244: we are using FxHashMap, which are faster than normal hash maps.

Eikix (approved):
lgtm

PR #563, opened 2023-10-26
Skip tests for local storage refacto

ClementWalter:


crates/ef-testing/src/models/case.rs
```diff
@@ -165,10 +169,15 @@ impl BlockchainTestCase {
...
+                let balance_diff = format!(
                     "balance mismatch for {:#20x}: expected {:#32x}, got {:#32x}",
                     address, expected_state.balance.0, actual
-                )));
+                );
+                diff.push(balance_diff);
+            }
+
+            if !diff.is_empty() {
+                return Err(RunnerError::Other(diff.join("\\n")));
```
greged93, line 180: I think we can remove this if it doesn't work

greged93 (commented):
One comment

crates/ef-testing/src/models/case.rs
ClementWalter: but I still need to input a string to `Other` and `diff` is a `Vec<String>`

greged93: Will be fixed in another PR

Issue #568, opened 2023-10-30
dev: improve compilation time for tests generation


greged93:
Right now, a procedural macro walks through the ef-tests folder and generates all tests. This is however quite slow and can take up to 3 minutes for compilation whenever a change is made to the test filter or the rust code. 

This can be improved by making a `build.rs` script, change the way threads are created to avoid bottlenecks and use a "cached" filter in order to only update test files that need to be updated.

PR #570, opened 2023-10-30
dev: accelerate tests

greged93:
# Pull Request type

Time spent on this PR: 2 days

Resolves: #568 

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [x] Testing

## What is the new behavior?
Instead of using a macro and storing all the generated tests in one file, the tests are separated into the same folder structure as the ef-tests data using a build script. Additionally, `par_iter` from `rayon` crate is moved from folder iteration to test case iteration, which lowers the change for bottlenecks creation and speeds up compilation time.

- [x] Yes
- [ ] No


crates/build-utils/src/filter.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub fn diff(&self, rhs: &Self) -> Vec<Folder> {
+        let mut diff = Vec::new();
+        diff.append(&mut map_diff(&self.filename, &rhs.filename));
+        diff.append(&mut map_diff(&self.regex, &rhs.regex));
+        diff.append(&mut map_diff(&self.test_name, &rhs.test_name));
+        diff
+    }
+}
+
+fn map_diff(lhs: &FilterMap, rhs: &FilterMap) -> Vec<Folder> {
```
ClementWalter, line 72: what is lhs and rhs?

crates/ef-testing/build.rs
```diff
@@ -0,0 +1,69 @@
...
+lazy_static! {
+    static ref SUITE_PATH: PathBuf = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("./ethereum-tests/BlockchainTests/GeneralStateTests");
+    static ref INTEGRATION_TESTS_PATH: PathBuf =
+        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("./tests/");
+}
+
+fn main() {
+    // Tell Cargo that if the given file changes -> to rerun this build script.
+    println!("cargo:rerun-if-changed={SKIPPED_TESTS}");
```
ClementWalter, line 22: maybe "cleaner" to use the macro directly

`rerun_if_changed!(SKIPPED_TESTS)`

https://docs.rs/cargo-emit/latest/cargo_emit/macro.rerun_if_changed.html

ClementWalter (dismissed):
minor suggestions

crates/build-utils/src/filter.rs
greged93: left hand side and right hand side

crates/ef-testing/build.rs
greged93: ah interesting, did not know you could use the macro

crates/build-utils/src/filter.rs
greged93: hahaha I have no clue what latin hypercube sampling is, but lhs and rhs is quite common in rust when doing [operations](https://doc.rust-lang.org/std/ops/trait.Mul.html).

ClementWalter (approved):

ClementWalter (approved):

PR #590, opened 2023-11-22
feat: add tests that fail because of gas

Eikix:
# Pull Request type

Time spent on this PR: 1day 
 
Took me some time but I finally understand  ef-tests better ! 

all these tests are skipped because of OOG


greged93 (approved):
lgtm!

PR #610, opened 2023-12-05
Skip some false positive raising maximum offset value exceeded

ClementWalter:
Need to skip for https://github.com/kkrt-labs/kakarot/pull/840 fixing some of these errors


greged93 (approved):
lgtm

PR #609, opened 2023-12-04
feat: use lazy static and clone to init sequencer

greged93:
# Pull Request type

Time spent on this PR: 0.1 day

Resolves: #NA

Please check the type of change your PR introduces:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Documentation content changes
- [ ] Testing

## Does this introduce a breaking change?

- [ ] Yes
- [x] No

PR #611, opened 2023-12-05
Dev/limit ci

greged93:
# Pull Request type

Time spent on this PR: 1 day

Resolves: #NA

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [x] Testing

## What is the new behavior?
Run ef-tests on ssj. Reduce the amount of ran tests to about 230 tests (only VmArithmeticTest) and leave responsibility to run full range of tests to the `kkrt` and `ssj` repos.

- [x] Yes
- [ ] No


.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   test:
     # trunk-ignore(actionlint/runner-label)
-    runs-on: ubuntu-latest-32-cores
+    runs-on: ubuntu-latest-16-cores
```
ClementWalter, line 15: can't we just use ubuntu-latest then?

crates/ef-testing/src/models/result.rs
```diff
@@ -2,35 +2,40 @@ use blockifier::{
...
+                #[cfg(feature = "v0")]
                 if let Some(call) = info.execute_call_info {
+                    use starknet::core::types::FieldElement;
+                    use starknet_api::hash::StarkFelt;
                     let events = get_kakarot_execution_events(&call);
                     // Check only one execution event.
                     if events.len() != 1 {
                         warn!(
-                            "{}::{} failed to find the single execution event: {:?}",
+                            "{}::{} version v0 failed to find the single execution event: {:?}",
```
ClementWalter, line 38: you can use `{version}` so that when it's updated in v1 there is no need to update this. And to keep the version name at one single place

ClementWalter (commented):
some comments

.github/workflows/test.yml
greged93: Think we can, 16 cores is running in 12 minutes right now so will probably run in 25 minutes, should be ok

crates/ef-testing/src/evm_sequencer/evm_state/v0.rs
```diff
@@ -25,6 +25,11 @@ use crate::evm_sequencer::utils::{
 };
 
 impl Evm for KakarotSequencer {
+    /// Sets up the evm state (coinbase, block number, etc.)
+    fn setup_state(&mut self) -> StateResult<()> {
```
Eikix, line 29: what is this for?

Eikix: ah ok! 
you need a specific fn for v1/v0 to setup differently,
we set a coinbase in ssj, whereas in kkt0 we don't,

for now in kkt0, this doesn't do anything

Eikix: can this be done with lazy_static like in #609 at the moment of init?

greged93: I can rename but I think having something that sets up the "state" (so block, timestamp, coinbase, gas_price,...) makes sense.

Eikix: I can agree that technically setting a coinbase is an "environment" responsibility,
But also, the coinbase "should" to be in the registry, so it's part of the kakarot system

greged93: not sure I understand, the registry is set during the setup of each individual account, in the `setup_account` function

greged93: what I could do is: move all storage setting into `setup_state` and only leave in the `lazy_static` what touches the contract classes.

Eikix (dismissed):

crates/ef-testing/src/models/result.rs
```diff
@@ -18,7 +18,10 @@ pub(crate) fn log_execution_result(
                 warn!("{} reverted:\n{}", case, err.replace("\\n", "\n"));
             } else {
                 info!("{} passed: {:?}", case, info.actual_resources);
+                #[cfg(feature = "v0")]
```
ClementWalter, line 21: we're supposed to have this in kkt1 as well though

ClementWalter (approved):

crates/ef-testing/src/models/result.rs
greged93: we do but currently the events implementations don't match, should I leave this and merge as is and wait for v1 update or do you want to wait?

PR #614, opened 2023-12-06
dev: cleaning

greged93:
# Pull Request type

Resolves: #NA

- [x] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no API changes)
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?
Refactors some parts of the code by:
- Renaming
- Removing unnecessary traits or error variants
- Add helpful macro for starknet storage 
- Improves the API for some of the types

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/build-utils/src/dir_reader.rs
+++ b/crates/build-utils/src/dir_reader.rs
@@ -1,6 +1,10 @@
 use std::collections::BTreeMap;
 use std::path::PathBuf;
+use std::sync::Arc;
+use std::sync::Mutex;
 
+use rayon::iter::IntoParallelRefIterator;
+use rayon::iter::ParallelIterator;
 use walkdir::DirEntry;
 use walkdir::WalkDir;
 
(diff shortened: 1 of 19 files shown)


crates/build-utils/src/dir_reader.rs
```diff
@@ -37,34 +41,45 @@ impl<'a> DirReader<'a> {
...
-            let full_path = entry.path();
-            if let Some(target) = &self.target {
-                if target.iter().any(|t| {
-                    full_path
-                        .ancestors()
-                        .any(|a| a.to_str().map(|s| s == t).unwrap_or_default())
-                }) {
-                    continue;
+    ) -> Result<(), eyre::Error> {
+        let target = self.target.clone();
```
Eikix, line 59: Why do we need to go through this?

The fact that previous implementation was passed by value was a problem?

crates/ef-testing/src/evm_sequencer/account/mod.rs
```diff
@@ -9,7 +9,30 @@ use starknet_api::{
     state::StorageKey,
 };
 
+#[macro_export]
+macro_rules! starknet_storage {
```
Eikix, line 13: beautiful

crates/ef-testing/src/evm_sequencer/evm_state/v1.rs
```diff
@@ -259,8 +259,8 @@ pub(crate) fn compute_storage_base_address(
     StorageKey(PatriciaKey::try_from(key_floored).unwrap()) // infallible
 }
 
-pub(crate) fn offset_storage_base_address(base_address: StorageKey, offset: i64) -> StorageKey {
-    let base_address = stark_felt_to_felt(*base_address.0.key()) + Felt252::from(offset);
+pub(crate) fn offset_storage_key(key: StorageKey, offset: i64) -> StorageKey {
```
Eikix, line 262: feels weird that this is semantically an addition, and that we somehow cannot offset directly the storageKey

greged93: I know :/ the API on `StorageKey` is pretty bad, they should just implement something like `Add`, would make it a lot easier to offset, maybe PR on starknet api?

crates/build-utils/src/dir_reader.rs
greged93: This is because I now updated the implementation to iter the folders in parallel with `par_iter`, which means I need to place `self` in an `Arc<Mutex>`, which makes me lose the ownership of `self`. 
This makes it that I can't just do `self.target` anymore in the `map`, so I need to `clone` target. I realized that `target` is actually an Option which implements `Copy` so I removed the clone :)

Eikix (approved):
lgtm

PR #672, opened 2024-03-06
feat: check run output gas_used

enitrat:
# Pull Request type

Time spent on this PR: 0.3d

Resolves: #{Issue number}

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?

Uses the output from the run to retrieve the `gas_used` value, instead of using the value from the block header.

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -90,7 +90,7 @@ impl BlockchainTestCase {
             RunnerError::Other(vec!["No transaction in pre state block".into()].into())
         })?;
 
-        tx_signed.transaction.set_chain_id(*CHAIN_ID);
+        tx_signed.transaction.set_chain_id(CHAIN_ID);
         let signature = sign_message(self.secret_key, tx_signed.signature_hash())
             .map_err(|err| RunnerError::Other(vec![err.to_string()].into()))?;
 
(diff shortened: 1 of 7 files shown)


crates/ef-testing/src/models/result.rs
```diff
@@ -5,75 +5,108 @@ use blockifier::{
         objects::{TransactionExecutionInfo, TransactionExecutionResult},
     },
 };
+use eyre::{eyre, Result};
```
greged93, line 8: Can we use cainome for this ?

crates/ef-testing/src/models/case.rs
```diff
@@ -97,35 +97,37 @@ impl BlockchainTestCase {
...
+        });
 
         // Get gas_used and base_fee from RLP block - as in some cases, the block header is not present in the test data.
         let sealed_block = SealedBlock::decode(&mut self.block.rlp.as_ref())
             .map_err(RunnerError::RlpDecodeError)?;
         let sealed_header = sealed_block.header.unseal();
 
         let coinbase = sealed_header.beneficiary;
-        let gas_used: U256 = U256::from(sealed_header.gas_used);
+        let gas_used: U256 = U256::from(output.gas_used);
```
greged93, line 130: I'm not sure why we do this: I think we should compare the two values, not just replace one with the other

enitrat: Used `expected` and `actual` values, compared them and pushed in `diff` message if not coherent.

greged93 (approved):

greged93 (commented):
just checking

PR #674, opened 2024-03-11
chore: show RunResources in test-skip.yml, print revert reason in test

enitrat:
# Pull Request type

Time spent on this PR: 0.1d

Resolves: #{Issue number}

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?

- Adapted utils to generate skipfile with failure reason in comment if it's a runresources error
- If a test fails and there's a revert reason - print it.

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -161,12 +161,12 @@ impl BlockchainTestCase {
         let post_state = self.post.clone().expect("Post state not found");
         let post_state = update_post_state(post_state, self.pre.clone());
 
-        let mut diff: Vec<String> = vec![];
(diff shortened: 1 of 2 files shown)


crates/ef-testing/src/models/case.rs
```diff
@@ -230,6 +230,9 @@ impl BlockchainTestCase {
         }
 
         if !diff.is_empty() {
+            if let Ok(revert_reason) = maybe_revert_reason {
+                diff.push(format!("revert reason: {}", revert_reason));
+            }
```
greged93, line 235: can you rename `diff` to translate the fact that it contains an accumulation of the errors encountered in the post state?

PR #677, opened 2024-03-12
feat: initial support dencun

### LINKED ISSUE #676: feat: support Cancun migration
## Feature Request

In the [last release](https://github.com/ethereum/tests/releases/tag/v13.1) of the test suite, Pyspec tests folder moved from `/src/BlockchainTestsFiller/Pyspecs` to `/src/GeneralStateTestsFiller/Pyspecs`. As such, they are now comprised in our test generation component which reads tests from the `GeneralState` directory.

However, this is not compatible with the way our runner works for a few reasons:
1. The generated functions from these tests are not formatted as `test_name_dxgxvx_ForkName`, but as `test_src/GeneralStateTestsFillerFiller/Pyspecs/berlin/eip2930_access_list/test_acl.py::test_access_list[fork_Cancun_minus_blockchain_test]()` which is not a valide name for a rust function. As such, we must think of a strategy to adapt these tests.

2. These .py tests now can have both state and blockchain tests. (Not sure yet if we need to run both)

3. (might be related to 2.) Among the generated `.json` files from python specification, some of them are present in `BlockchainTests/GeneralStateTests/Pyspects/cancun/test_folder/test_name.json` but not in `

4. These Pyspec tests might be duplicated from some other tests - for example, `crates/ef-testing/ethereum-tests/src/BlockchainTestsFiller/Pyspecs/shanghai/eip3855_push0/test_push0.py` is a duplicate of `src/GeneralStateTestsFiller/VMTests/vmTests/push0Filler.yml`. Considering this, we must evaluate the need to run the tests generated from the python specification. If it is not required, we can simply skip them.

One of these tests is `crates/ef-testing/ethereum-tests/BlockchainTests/GeneralStateTests/Pyspecs/cancun/eip4788_beacon_root/multi_block_beacon_root_timestamp_calls.json`, which doesn't exist under `crates/ef-testing/ethereum-tests/GeneralStateTests/Pyspecs/cancun/eip4788_beacon_root`. As such, the `secret_key` function fails.

https://github.com/kkrt-labs/ef-tests/blob/6fefabc997a434733e6fa915957ea02e8fcd872c/crates/build-utils/src/content_reader.rs#L25-L37

- [ ] #681
- [ ] #682
- [x] #683
```[tasklist]
### Tasks
- [ ] https://github.com/kkrt-labs/ef-tests/issues/683
- [ ] https://github.com/kkrt-labs/ef-tests/issues/682
- [ ] https://github.com/kkrt-labs/ef-tests/issues/681
- [ ] https://github.com/kkrt-labs/ef-tests/issues/684
```

### LINKED ISSUE #681: Ignore the `beacon_root` post-state validations

### LINKED ISSUE #682: Add a fallback mechanism to get the secretKey of Pyspecs tests

### LINKED ISSUE #683: Adapt test name generation to support Pyspecs tests
Pyspecs-based tests have a different structure than other ones.

1. They're defined in `Pyspecs/cancun/folder_name/tested_feature.py`
2. Inside `tested_feature.py`, there are multiple tests defined.
3. These tests contain unsupported chars that clash with the rust test name generation that must be replaced

enitrat:
# Pull Request type

Time spent on this PR: 2 days

Resolves: #684 (doesn't support block with txs from different senders as the `secretKey` is a single value for now)
Resolves #683 
Resolves #682
Resolves #681 
Resolves #676 

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

- [ ] Yes
- [ ] No


crates/build-utils/src/content_reader.rs
```diff
@@ -24,7 +24,11 @@ impl ContentReader {
     /// Secret key location: GeneralStateTests/stRandom/randomStatetest0.json
     pub fn secret_key(path: PathWrapper) -> Result<Option<Value>, eyre::Error> {
         let path = blockchain_tests_to_general_state_tests_path(path);
-        let content = path.read_file_to_string()?;
+        let maybe_content = path.read_file_to_string();
```
enitrat, line 27: There might not be a corresponding file in the general state path with the format expected (only for Pyspecs cases)

crates/build-utils/src/content_reader.rs
```diff
@@ -71,4 +75,31 @@ impl ContentReader {
         // Return a clone of the block
         Ok(first_block.clone())
     }
+
+    pub fn transaction(test_case: &Value, block: &Value) -> Result<Value, eyre::Error> {
```
enitrat, line 79: goal: get the tx sender later on and use it to get it's secret key from a known mapping

greged93, line 79: From what I could see we only use `ContentReader::transaction` in order to get the private key. Could we perhaps just move this logic into the private key function? We can then move the whole private key logic in `format_test_content`, pass the block and get the pvt key using the static only if needed

greged93 (commented):
just one comment but looks nice otherwise

crates/build-utils/src/content_reader.rs
```diff
@@ -71,4 +100,31 @@ impl ContentReader {
...
```
greged93, line 104: we can delete this then?

crates/ef-testing/Cargo.toml
```diff
@@ -8,6 +8,7 @@ repository.workspace = true
 build = "build.rs"
 
 [dependencies]
+build-utils = { path = "../build-utils" }
```
greged93, line 11: are we using this?

scripts/generate_skip_file.py
```diff
@@ -42,20 +54,21 @@ def parse_and_write_to_yaml(input_file, output_file):
...
             .replace("_xor_", "^"),
             m.split("::")[-1]
-            .replace("_minus_", "-")
-            .replace("_plus_", "+")
-            .replace("_xor_", "^"),
         )
         for m in re.findall(r"thread '(.*)' panicked at", data)
     ]
 
+    print(matches_failed)
```
greged93, line 61: remove

scripts/generate_skip_file.py
```diff
@@ -42,20 +54,21 @@ def parse_and_write_to_yaml(input_file, output_file):
...
+
     summary = next(
         re.finditer(
             r"test result: (?P<result>\w+). (?P<passed>\d+) passed; (?P<failed>\d+) failed; (?P<ignored>\d+) ignored",
             data,
         )
     )
 
+    print(summary)
```
greged93, line 70: remove

crates/build-utils/src/content_reader.rs
enitrat: we still use it inside secret_key to get the tx from a block

crates/ef-testing/Cargo.toml
enitrat: actually, it can stay as a build util

greged93: I'm not sure I understand, do you need to import this in ef-testing?

enitrat: yes - but it was already importer as a build-dependency so I will leave it as one

greged93 (approved):
lgtm

PR #678, opened 2024-03-16
fix: update lock

greged93:
# Pull Request type

Time spent on this PR: 0.1 day

Resolves: #NA

- [x] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?
Updates the lock to point to the new commits of the cairo vm and blockifier forks.

- [ ] Yes
- [x] No

### DIFF
--- a/crates/ef-testing/src/models/case.rs
+++ b/crates/ef-testing/src/models/case.rs
@@ -156,7 +156,6 @@ impl BlockchainTestCase {
             ));
         }
         let gas_price = gas_price | effective_gas_price;
-        let transaction_cost = gas_price * expected_gas_used;
 
         let post_state = self.post.clone().expect("Post state not found");
         let post_state = update_post_state(post_state, self.pre.clone());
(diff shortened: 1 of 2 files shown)


crates/ef-testing/src/models/case.rs
```diff
@@ -156,7 +156,6 @@ impl BlockchainTestCase {
             ));
         }
         let gas_price = gas_price | effective_gas_price;
-        let transaction_cost = gas_price * expected_gas_used;
```
Eikix, line 159: we don't need this anymore ?

Eikix (commented):
one question

crates/ef-testing/src/models/case.rs
greged93: No this isn't used anymore, was removed in a PR by eni

Eikix (approved):
lgtm

PR #731, opened 2024-07-29
feat: set code hash in pre state

obatirou:
# Pull Request type

Linked to https://github.com/kkrt-labs/kakarot/pull/1309#issuecomment-2255276974

- [ ] Bugfix
- [x] Feature
- [ ] Build-related changes
- [ ] Testing

## What is the new behavior?

Code hash is set in storage when handling the pre state 

- [ ] Yes
- [x] No

### DIFF
--- a/crates/ef-testing/src/evm_sequencer/account/v0.rs
+++ b/crates/ef-testing/src/evm_sequencer/account/v0.rs
@@ -65,7 +65,7 @@ impl KakarotAccount {
         let code_hash_values = split_u256(code_hash);
         let code_hash_low_key = get_storage_var_address(ACCOUNT_CODE_HASH, &[]);
         let code_hash_high_key = next_storage_key(&code_hash_low_key)?;
-        storage.append(&mut vec![
+        storage.extend([
             (code_hash_low_key, Felt::from(code_hash_values[0])),
             (code_hash_high_key, Felt::from(code_hash_values[1])),
         ]);
--- a/crates/ef-testing/src/evm_sequencer/account/v0.rs
+++ b/crates/ef-testing/src/evm_sequencer/account/v0.rs
@@ -1,5 +1,8 @@
 use blockifier::abi::{abi_utils::get_storage_var_address, sierra_types::next_storage_key};
+use reth_primitives::alloy_primitives::keccak256;
+use reth_primitives::KECCAK_EMPTY;
 use reth_primitives::{Address, U256};
+
 use revm_interpreter::analysis::to_analysed;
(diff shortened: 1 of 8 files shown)


crates/ef-testing/src/models/case.rs
```diff
@@ -65,6 +65,12 @@ impl BlockchainTestCase {
             )?;
             sequencer.setup_account(kakarot_account)?;
             sequencer.fund(address, account.balance)?;
+            if account.balance != U256::from(0)
```
obatirou, line 68: Do we want to put this behind a feature flag or do we want to duplicate the set_code_hash function in V1 ?
I would say the latter but maybe put it in default implementation of the Evm trait instead of duplication ?

enitrat: I would rather have the account code hash set in `KakarotAccount::new` in which you can also pass the balance so that you can trigger the condition has_code_or_nonce_or_balance to know whether the hash is h(bytecode) or h(empty) or 0

obatirou: It might be better indeed. But now the v1 will need to accept the balance parameter and not use it.

obatirou: done with [591ae5e](https://github.com/kkrt-labs/ef-tests/pull/731/commits/591ae5e90a598d12f075ab3a49d31e2cd4760763)

obatirou:
Cannot be merged until a release is done in Kakarot

crates/ef-testing/src/evm_sequencer/account/v0.rs
```diff
@@ -47,6 +51,25 @@ impl KakarotAccount {
...
+            U256::from_be_slice(keccak256(code).as_slice())
+        };
+
+        let code_hash_values = split_u256(code_hash);
+        let code_hash_low_key = get_storage_var_address(ACCOUNT_CODE_HASH, &[]);
+        let code_hash_high_key = next_storage_key(&code_hash_low_key)?;
+        storage.append(&mut vec![
+            (code_hash_low_key, Felt::from(code_hash_values[0])),
+            (code_hash_high_key, Felt::from(code_hash_values[1])),
+        ]);
```
tcoratger, line 71: Here I think you could use something like:

```rust
storage.extend([
    (code_hash_low_key, Felt::from(code_hash_values[0])),
    (code_hash_high_key, Felt::from(code_hash_values[1])),
]);
```

enitrat (approved):

PR #737, opened 2024-08-20
fix: ssj can run again

enitrat:
# Pull Request type

Time spent on this PR:

Resolves: #{Issue number}

- [ ] Bugfix
- [ ] Feature
- [ ] Build-related changes
- [ ] Testing

- [ ] Yes
- [ ] No

### DIFF
--- a/crates/ef-testing/src/evm_sequencer/account/v1.rs
+++ b/crates/ef-testing/src/evm_sequencer/account/v1.rs
@@ -84,7 +84,7 @@ impl KakarotAccount {
         };
 
         let jumdpests_storage_address = get_storage_var_address(ACCOUNT_VALID_JUMPDESTS, &[]);
-        let jumdpests_storage_address = *jumdpests_storage_address.0.key();
+        let jumdpests_storage_address = Felt::from(jumdpests_storage_address);
         valid_jumpdests.into_iter().for_each(|index| {
             storage.push((
                 (jumdpests_storage_address + Felt::from(index))
--- a/crates/ef-testing/src/evm_sequencer/account/v1.rs
+++ b/crates/ef-testing/src/evm_sequencer/account/v1.rs
@@ -1,56 +1,28 @@
 use blockifier::abi::abi_utils::get_storage_var_address;
-use reth_primitives::{Address, Bytes, U256};
+use blockifier::abi::sierra_types::next_storage_key;
+use reth_primitives::{keccak256, Address, Bytes, KECCAK_EMPTY, U256};
+use revm_interpreter::analysis::to_analysed;
+use revm_primitives::Bytecode;
 use starknet::core::types::Felt;
 use starknet_api::{core::Nonce, state::StorageKey, StarknetApiError};
 
 use super::KakarotAccount;
 use super::{inner_byte_array_pointer, pack_byte_array_to_starkfelt_array};
 use crate::evm_sequencer::constants::storage_variables::{
-    ACCOUNT_EVM_ADDRESS, ACCOUNT_IS_INITIALIZED, ACCOUNT_NONCE, ACCOUNT_STORAGE,
+    ACCOUNT_BYTECODE_LEN, ACCOUNT_CODE_HASH, ACCOUNT_EVM_ADDRESS, ACCOUNT_IS_INITIALIZED,
+    ACCOUNT_NONCE, ACCOUNT_STORAGE, ACCOUNT_VALID_JUMPDESTS,
 };
 use crate::evm_sequencer::{
     constants::storage_variables::ACCOUNT_BYTECODE, types::felt::FeltSequencer, utils::split_u256,
 };
-use crate::{
-    evm_sequencer::evm_state::v1::{compute_storage_base_address, offset_storage_key},
-    starknet_storage,
-};
-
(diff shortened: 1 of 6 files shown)


crates/ef-testing/src/evm_sequencer/account/v1.rs
```diff
@@ -66,25 +38,70 @@ impl KakarotAccount {
...
+            .enumerate()
+            .map(|(i, bytes)| (StorageKey::from(i as u32), bytes))
+            .collect();
         storage.append(&mut bytecode_storage);
 
+        // Initialize the code hash var
+        let account_is_empty =
+            code.is_empty() && nonce == Felt::from(0) && balance == U256::from(0);
+        let code_hash = if account_is_empty {
+            U256::from(0)
```
greged93, line 58: Don't think this is ever valid, empty account should be keccak empty too I believe

greged93 (changes requested):

crates/ef-testing/src/evm_sequencer/account/v1.rs
```diff
@@ -1,56 +1,28 @@
...
-        })
-        .collect();
-    bytearray.extend(bytecode_storage);
-
-    bytearray
-}
+use crate::starknet_storage;
 
 impl KakarotAccount {
     pub fn new(
```
obatirou, line 21: If not mistaken this now completely identical to v0 
Should this be refactored to only have one file and avoid v0 / v1 ?

enitrat: eventually yes but before deleting everything I want to make sure I won't have other changes to make later

crates/ef-testing/src/evm_sequencer/account/v1.rs
enitrat: no, empty (non-existing) accounts codehash are zero, not empty keccak

greged93: Can you send spec? Yellow paper and reth use keccak empty

crates/ef-testing/src/evm_sequencer/account/v1.rs
```diff
@@ -66,25 +38,70 @@ impl KakarotAccount {
...
+                .0
+                .iter()
+                .enumerate()
+                .filter_map(|(index, bit)| bit.as_ref().then(|| index))
+                .collect(),
+            _ => unreachable!("Bytecode should be analysed"),
+        };
+
+        let jumdpests_storage_address = get_storage_var_address(ACCOUNT_VALID_JUMPDESTS, &[]);
+        let jumdpests_storage_address = *jumdpests_storage_address.0.key();
```
obatirou, line 87:
```suggestion
        let jumdpests_storage_address = Felt::from(jumdpests_storage_address);
```
I think this work, imo clearer

crates/ef-testing/src/evm_sequencer/evm_state/v1.rs
```diff
@@ -185,13 +187,8 @@ impl Evm for KakarotSequencer {
     fn nonce_at(&mut self, evm_address: &Address) -> StateResult<U256> {
```
obatirou, line 187: Same here, most function should be completely identical to v0 right ?

crates/ef-testing/src/evm_sequencer/account/v1.rs
obatirou: Ok good for me then

crates/ef-testing/src/evm_sequencer/account/v1.rs
obatirou: https://eips.ethereum.org/EIPS/eip-1052

greged93: Ok this is a problem for the rpc side I think, because extcode hash needs to return 0 but I don't think we should return 0 on the rpc (cfr screenshot)

image

enitrat: @greged93 It is possible that the yellow paper is not correctly updated after each merged EIP, it's a bit sad but it is not a reliable source of info when it comes to very specific details :/

greged93: No no the yellow paper is correct: the code hash is empty keccak. But the extcode hash returns 0 on dead contracts. It's just not the same afaiu

enitrat: ah, ok; but regardless it has happened multiple times that the yellow paper was not exactly matching execution-specs

greged93: I'll make a query to a geth node to check what it does just to be sure

enitrat: when in doubt -> https://github.com/ethereum/execution-specs/blob/07f5747a43d62ef7f203d41d77005cb15ca5e434/src/ethereum/cancun/vm/instructions/environment.py#L477-L481

obatirou (approved):

greged93 (approved):

PR #747, opened 2024-09-11
feat: remove distinctions between KakarotZero and KakarotSSJ

enitrat:
Removes most distinctions between the two codebases now that they're similar.
One thing remains - the fee charged is not (yet) verified in v1.


crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
```diff
@@ -197,3 +200,87 @@ pub fn convert_contract_class_v1(class: &CompiledClass) -> Result<ContractClass,
...
+        ];
+
+        #[cfg(not(feature = "v0"))]
+        let storage = vec![
+            (OWNABLE_OWNER, *KAKAROT_OWNER_ADDRESS.0.key()),
+            (KAKAROT_NATIVE_TOKEN_ADDRESS, *ETH_FEE_TOKEN_ADDRESS.0.key()),
+            (KAKAROT_ACCOUNT_CONTRACT_CLASS_HASH, ACCOUNT_CONTRACT_CLASS_HASH.0),
+            (KAKAROT_BLOCK_GAS_LIMIT, Felt::from(BLOCK_GAS_LIMIT)),
+            (KAKAROT_UNINITIALIZED_ACCOUNT_CLASS_HASH, UNINITIALIZED_ACCOUNT_CLASS_HASH.0),
+        ];
```
obatirou, line 226: for `v0` you could only append the `(KAKAROT_CAIRO1_HELPERS_CLASS_HASH, CAIRO1_HELPERS_CLASS_HASH.0)`

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
```diff
@@ -197,3 +200,87 @@ pub fn convert_contract_class_v1(class: &CompiledClass) -> Result<ContractClass,
...
+
+        // Write all the storage vars to the sequencer state.
+        for (k, v) in storage {
+            (&mut state).set_storage_at(*KAKAROT_ADDRESS, get_storage_var_address(k, &[]), v).expect("failed to set storage");
+        }
+
+        // Convert the contract classes to using the util corresponding to the Cairo version it's compiled with.
+        let (converted_kakarot_class, converted_account_class, converted_uninitialized_class) = {
+            #[cfg(not(feature = "v1"))]
```
obatirou, line 235: use `v0` and remove the negation ?

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
enitrat: yes - I tried that, but then the linter would complain of an unused `mut` in the case of `v1` - thus I preferred to split both. Idk what would be the correct approach here.

crates/ef-testing/src/evm_sequencer/constants.rs
```diff
@@ -72,13 +75,11 @@ lazy_static! {
...
-    pub static ref OPENZEPPELIN_ACCOUNT_CLASS: LegacyContractClass = load_contract_class("../../build/v0/OpenzeppelinAccount.json").expect("Failed to load openzeppelin account contract class");
 
     // Main class hashes
     pub static ref KAKAROT_CLASS_HASH: ClassHash = ClassHash(KAKAROT_CLASS.class_hash().unwrap());
     pub static ref ACCOUNT_CONTRACT_CLASS_HASH: ClassHash = ClassHash(ACCOUNT_CONTRACT_CLASS.class_hash().unwrap());
     pub static ref UNINITIALIZED_ACCOUNT_CLASS_HASH: ClassHash = ClassHash(UNINITIALIZED_ACCOUNT_CLASS.class_hash().unwrap());
-    pub static ref OPENZEPPELIN_ACCOUNT_CLASS_HASH: ClassHash = ClassHash(OPENZEPPELIN_ACCOUNT_CLASS.class_hash().unwrap());
 }
 
 #[cfg(feature = "v1")]
```
obatirou, line 85: still necessary ?
There is a flag only for v1 why ?

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
enitrat: causes an issue with clippy about the expected types for "default" targets - if I only have v0 and v1 target, the expected type is a (), whereas here it understands that in _all_ cases is returned a tuple of size 3

crates/ef-testing/src/evm_sequencer/constants.rs
enitrat: there are flags for both actually, to say where we take the classes from

crates/ef-testing/src/evm_sequencer/utils.rs
```diff
@@ -82,57 +82,54 @@ pub fn to_broadcasted_starknet_transaction(
...
+                    Felt::from(10_000_000_000_000u128), // OutsideExecution execute_before
+                    Felt::ONE,                          // call_array_len
+                    *KAKAROT_ADDRESS.0.key(),           // CallArray to
+                    selector!("eth_send_transaction"),  // CallArray selector
+                    calldata.len().into(),              // CallArray data_len
+                ];
+                execute_from_outside_calldata.append(&mut calldata);
+                execute_from_outside_calldata.push(signature.len().into());
+                execute_from_outside_calldata.append(&mut signature);
+                execute_from_outside_calldata
```
obatirou, line 117: this could be move outside the big if ?

crates/ef-testing/src/evm_sequencer/utils.rs
```diff
@@ -82,57 +82,54 @@ pub fn to_broadcasted_starknet_transaction(
...
+                    *RELAYER_ADDRESS.0.key(),           // OutsideExecution caller
+                    Felt::ZERO,                         // OutsideExecution nonce
+                    Felt::ZERO,                         // OutsideExecution execute_after
+                    Felt::from(10_000_000_000_000u128), // OutsideExecution execute_before
+                    Felt::ONE,                          // call_array_len
+                    *KAKAROT_ADDRESS.0.key(),           // CallArray to
+                    selector!("eth_send_transaction"),  // CallArray selector
+                    Felt::ZERO,                         // CallArray data_offset
+                    calldata.len().into(),              // CallArray data_len
+                    calldata.len().into(),              // calldata_len
```
obatirou, line 97: the only diff if this line
Maybe remove the if block and append only for v0 ?

crates/ef-testing/src/evm_sequencer/utils.rs
```diff
@@ -144,7 +141,7 @@ pub fn to_broadcasted_starknet_transaction(
                 Felt::ONE,                                   // version
                 relayer_address,                             // sender_address
                 Felt::ZERO,                                  // 0
-                compute_hash_on_elements(&execute_calldata), // h(calldata)
+                compute_hash_on_elements(&execute_entrypoint_calldata), // h(calldata)
```
obatirou, line 143: lint corresponding comments to be on the same column

crates/ef-testing/src/evm_sequencer/utils.rs
enitrat: I prefer to have it explicit here as it's not super straightforward - in two blocks the transaction structure is outlined better

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
obatirou: I see, there is already a `#[allow(unused_mut)` for the signature due to this problem in `to_broadcasted_starknet_transaction`.

IMO adding `unused_mut` + comment explaining why will simplify the code but @tcoratger might have a better sense of idiomatic rust

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
obatirou: maybe add a third case like the rest of the code base ?
Shouldn't it panics if not V0 or V1 ?

crates/ef-testing/src/evm_sequencer/constants.rs
obatirou: Oh ok missed this, thx

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
enitrat: I don't think I can because this must be done at compile time, not runtime here because of type differences

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
enitrat: ah good idea!

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
```diff
@@ -40,12 +38,20 @@ use starknet_api::{
...
+use crate::evm_sequencer::constants::{
+    storage_variables::KAKAROT_CAIRO1_HELPERS_CLASS_HASH, CAIRO1_HELPERS_CLASS,
+    CAIRO1_HELPERS_CLASS_HASH,
 };
 
+use blockifier::abi::abi_utils::get_storage_var_address;
+#[allow(unused_imports)]
+use blockifier::state::state_api::{
+    State as BlockifierState, StateReader as BlockifierStateReader,
+};
```
tcoratger:
```suggestion
#[cfg(feature = "v0")]
use crate::evm_sequencer::constants::{
    storage_variables::KAKAROT_CAIRO1_HELPERS_CLASS_HASH, CAIRO1_HELPERS_CLASS,
    CAIRO1_HELPERS_CLASS_HASH,
};
use blockifier::abi::abi_utils::get_storage_var_address;
#[allow(unused_imports)]
use blockifier::state::state_api::{
    State as BlockifierState, StateReader as BlockifierStateReader,
};
```

crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
```diff
@@ -1,26 +1,24 @@
...
-use lazy_static::lazy_static;
-#[cfg(not(any(feature = "v0", feature = "v1")))]
-lazy_static! {
-    pub static ref INITIAL_SEQUENCER_STATE: State = State::default();
-}
-
 use starknet::core::types::Felt;
 use std::ops::{Deref, DerefMut};
 
 use crate::evm_sequencer::types::felt::FeltSequencer;
```
tcoratger, line 5:
```suggestion
use std::ops::{Deref, DerefMut};
use crate::evm_sequencer::types::felt::FeltSequencer;
```

tcoratger (approved):

PR #752, opened 2024-09-12
refactor `KakarotSequencer` impl

tcoratger:
- Remove useless functions
- Refactor the implementation

### DIFF
--- a/crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
+++ b/crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
@@ -95,12 +95,10 @@ impl KakarotSequencer {
         block_number: u64,
         block_timestamp: u64,
     ) -> Self {
-        let coinbase_constructor_args = {
-            vec![
-                Felt::ONE,
-                Felt::from_bytes_be_slice(&coinbase_address.0[..]),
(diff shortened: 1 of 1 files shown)


crates/ef-testing/src/evm_sequencer/sequencer/mod.rs
```diff
@@ -162,11 +152,12 @@ impl KakarotSequencer {
     }
 
     pub fn compute_starknet_address(&self, evm_address: &Address) -> StateResult<ContractAddress> {
-        let base_class_hash = self.environment.base_account_class_hash.0;
-
-        let constructor_args = { vec![Felt::ONE, Felt::from_bytes_be_slice(&evm_address.0[..])] };
-
-        Ok(compute_starknet_address(evm_address, base_class_hash, &constructor_args).try_into()?)
+        Ok(compute_starknet_address(
```
greged93, line 155: I think I prefer the version where we explicitly name the intermediate variables as it translates more what we are doing

tcoratger: Sure just reverted but I also removed the useless usage of vec here

greged93 (approved):

